Issue #177, opened 2014-04-28
bug in Datalog expressions


stechu:
``` datalog
A(x) :- R(x,3), x>3+1.
```

will compile to a wrong RA plan and a wrong Myria plan. The expression `3+1` will be ignored.
Is this by design or is this a bug?

Issue #107, opened 2014-02-07
Datalog parser should fail on bad input


bmyerz:
In some cases the parser quietly drops input when syntax is not supported

```
A(x) :- R(x), x<3 or x>10
```

The parser should fail rather than quietly producing:

```
A = Project($0)[Select(($0 < 3))[Scan(public:adhoc:R)]]
```

PR #180, opened 2014-04-28
[OPEN WIP] Fix datalog bugs

dhalperi:
Catch-all PR for Datalog bugs.

1. Please add tests for any existing Datalog bugs here. I caught #107 and #177 so far. The expectation is that the tests should fail!

2. Please fix the bugs. Once we have a comprehensive test suite and all the bugs are fixed, we can be more confident in Datalog working.


dhalperi:
Closing for inactivity.

PR #254, opened 2014-06-19
Add aql escape syntax to myrial and raco

7andrew7:
Syntax like the following is supported:

```
%aql select * from MyArray;
X = Scan() -- Normal Myrial code
```


dhalperi:
Closing for inactivity

PR #268, opened 2014-07-09
[WIP] fix 138

### LINKED ISSUE #138: MyriaL can emit plans with needless cross products
Myrial / raco doesn't reorder joins: the order specified by the user is the final join order.  This can lead to plans with needless cross-products in the event that the user specified two adjacent relations that don't participate in a join.

For example, this follow plan might emit a needless cross product of A and B:

```
X = [FROM A, B, C WHERE A.x == C.x and B.y == C.y EMIT *];
```

stechu:
fix #138


dhalperi:
Closing for inactivity

PR #323, opened 2014-09-09
Store schema rename

TheYorkWei:
Appends .schema to the schema files written by CStore and GrappaStore
Adds a try/except to run_query.

- merge after PR #359


dhalperi:
@bmyerz any idea what's going on with this? for now, closing for inactivity as it is not even mergable any more.

bmyerz:
for posterity sake:
This branch only has minor tweaks to the clang backend that I have to go through again anyway to deploy it more robustly

PR #350, opened 2014-09-20
Avg stdev cast double

dhalperi:
Basically, I think we want the semantics of things like average to be defined by MyriaL rather than by the backends. That is more like the "write once, run anywhere" model we want for Raco.

As a start, make AVG and STDEV force-insert casts to doubles. However, this breaks the tests.

@7andrew7 pretty unhappy with this hack, any suggestions? I noticed you had a note that you did *not* want to force semantics in MyriaL, so I thought maybe I'd let you ring in..


bmyerz:
@7andrew7 @dhalperi what is the status of this issue? Can it be closed and marked as a discussion thread? Or should it be merged?

PR #360, opened 2014-10-31
Add apache license, fixes #278

### LINKED ISSUE #278: Does raco have an explicit license?
Does raco have an explicit license? I've talked with @domoritz and looked at the website, and understand it's meant to be released as open source, but can't find it in the repository.

domoritz:


stechu:
@dhalperi  @domoritz  are we using any third party non-apache compatible libraries in RACO?

dhalperi:
I think we're okay here.

Issue #362, opened 2014-11-11
Crazy idea: add unicode comparisons to MyriaL?


dhalperi:
- `≤` same as `<=`
- `≥` same as `>=`
- `≠` same as `!=`

what else?

Okay, this is silly but it might actually make programs more readable in the long run. And people will remember it.

(You can type these on a Mac by using `option` with `,`, `.`, and  `=`. That is, don't push the shift key in when doing `option` + `<` or `>`.)

PR #363, opened 2014-11-12
RFC: add Unicode math operators

dhalperi:
Thoughts related to the general idea should go on #362 ; comments on implementation fit here :)


stechu:
LGTM.

Issue #283, opened 2014-08-22
give user hints in error messages


dhalperi:
* [ ] Unexpected end-of-file - add (maybe you are missing a semicolon?)?

* [ ] Optimized program is empty - add (are you saving your output with have a store statement? store(variable, user:program:dataset);)

* [ ] Convert assert to reasonable error:
    ```myrial
    y = empty(a:int);
    z = [from y s -- bug: errant s
         emit y.*];
    store(z, debug);
    ```
    currently yields
    ```python
    File "/base/data/home/apps/s~myria-web/1.381369730358170872/raco/myrial/emitarg.py", line 101, in expand_relation
      assert relation_name in symbols
    AssertionError
    ```

* [ ] Another inscrutable parser error on very slightly different query:
    ```myrial
    y = empty(a:int);
    z = [from y s -- bug: extra s
         emit y.a];
    store(z, debug);
    ```
    yields
    ```python
    File "/base/data/home/apps/s~myria-web/1.381369730358170872/raco/myrial/interpreter.py", line 76, in evaluate
      method = getattr(self, expr[0].lower())
    AttributeError: 'ExpressionProcessor' object has no attribute 'y'
    ```

PR #382, opened 2015-01-08
[WIP] Error messages

dhalperi:
See #283

### DIFF
--- a/raco/myrial/interpreter.py
+++ b/raco/myrial/interpreter.py
@@ -143,8 +143,18 @@ class ExpressionProcessor(object):
             if isinstance(sub_expr, raco.expression.Unbox):
                 rex = sub_expr.relational_expression
                 if rex not in from_args:
-                    unbox_op = self.evaluate(rex)
-                    from_args[rex] = unbox_op
+                    # TODO this check seems very strange.
+                    #  1) rex is a string -- look up its symbol above
+                    #  2) rex is not in that table, so it's an alias or other
+                    #     expression. Evaluate it.
+                    #  which missed 3) rex is a string but not in the table,
+                    #     it's a typo. Evaluating leads to weird error, so
+                    #     make sure it's an expression by proxy to it's a tuple
+                    if type(rex) is tuple:
+                        unbox_op = self.evaluate(rex)
+                        from_args[rex] = unbox_op
+                    else:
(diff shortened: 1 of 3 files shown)


dhalperi:
@7andrew7 might we chat about this briefly tomorrow? Been trying to the bugs myself to ensure that I can keep the parser up, and have some design questions. See the TODO in commit 2fd9008.

Some questions about rules. E.g., for [Unbox](https://github.com/uwescience/raco/blob/master/raco/myrial/parser.py#L818),

```python
    @staticmethod
    def p_sexpr_id_dot_id(p):
        'sexpr : unreserved_id DOT unreserved_id'
        p[0] = sexpr.Unbox(p[1], p[3])

    @staticmethod
    def p_sexpr_id_dot_pos(p):
        'sexpr : unreserved_id DOT DOLLAR INTEGER_LITERAL'
        p[0] = sexpr.Unbox(p[1], p[4])
```

why are these two variants explicitly spelled out instead of using the [`column_ref` expression](https://github.com/uwescience/raco/blob/master/raco/myrial/parser.py#L768).

Second, more germane to this bug, why isn't there an `alias` expression that we used in place of the `unreserved_id` in the two definitions here, as well as where the `('ALIAS', ..)` tuple is created: https://github.com/uwescience/raco/blob/master/raco/myrial/parser.py#L768 . Then we wouldn't need to handle the two different types of `rex` in the code in that fix.

(I will fix the bugs myself, but would appreciate your advice!)

dhalperi:
I updated the `column_ref` redundancy in c643da3, and that worked fine.

In a different branch, I had my idea for using `ALIAS`, but I clearly was misguided. I think an Alias means something different than I was thinking. (see c208f18 in another branch, which is buggy.)

7andrew7:
I think this is a parser bug -- i.e., the code should generate a parse error instead of some sort of relation not found error.  I'm cooking up an alternate fix...

7andrew7:
As you probably deduced, the logic here was very convoluted.  I tried to simplify things in #383.

Conceptually, every table in a from clause has a name (a string) and a corresponding Operator.  The previous Unboxing implementation was hideous in that it allowed the name to sometimes be a tuple containing an AST.  Let me know if #383 clarifies the situation.

PR #383, opened 2015-01-09
Simplify design, implementation of unboxing in Myrial

### LINKED ISSUE #283: give user hints in error messages

7andrew7:
This PR simplifies the implementation of unboxing in Myrial.  More controversially, it also simplifies the language-level design of this feature.

At the language level, we now disallow non-trivial unbox expressions.  That is, you can only unbox a table name, not some arbitrary expression.  Previously, you could say this:

```python
 out = [FROM emp WHERE id > *COUNTALL(dept) EMIT emp.id];
```

Now, you must use this:

```python
out = [FROM emp, COUNTALL(dept) as size WHERE id > *size EMIT emp.id];
```

This simplifies the overall design, since unbox expressions are now just simple column references.

I clarified the implementation by distinguishing DottedRef expressions (e.g., MyTable.MyColumn) from Unbox expressions ($SomeTable.MyColumn).  I also added various comments and code cleanup.

Fixes #283 
Subsumes #382

### DIFF
--- a/examples/reachable.myl
+++ b/examples/reachable.myl
@@ -9,6 +9,6 @@ DO
                               EMIT Edge.dst AS addr]);
     Delta = DIFF(NewlyReachable, Reachable);
     Reachable = UNIONALL(Delta, Reachable);
-WHILE [*COUNTALL(Delta) > 0];
+WHILE [FROM COUNTALL(Delta) AS size EMIT *size > 0];
 
 STORE(Reachable, OUTPUT);
--- a/raco/myrial/emitarg.py
+++ b/raco/myrial/emitarg.py
@@ -47,7 +47,7 @@ def resolve_unbox(sx, symbols):
         return sx.field
     else:
         assert isinstance(sx.field, int)
-        op = symbols[sx.relational_expression]
+        op = symbols[sx.table_name]
         scheme = op.scheme()
         return scheme.getName(sx.field)
 
--- a/raco/myrial/emitarg.py
+++ b/raco/myrial/emitarg.py
@@ -143,4 +143,3 @@ class FullWildcardEmitArg(EmitArg):
 
     def __repr__(self):
         return 'FullWildcardEmitArg()'
-
--- a/raco/myrial/emitarg.py
+++ b/raco/myrial/emitarg.py
@@ -124,7 +124,7 @@ class TableWildcardEmitArg(EmitArg):
         return expand_relation(self.relation_name, symbols)
 
     def __repr__(self):
-        return 'TableWildcardEmitArg(%s)' % self.relation_name
+        return 'TableWildcardEmitArg(%r)' % self.relation_name
 
 
 class FullWildcardEmitArg(EmitArg):
--- a/raco/myrial/emitarg.py
+++ b/raco/myrial/emitarg.py
@@ -95,6 +95,9 @@ class NaryEmitArg(EmitArg):
     def get_statemods(self):
         return self.statemods
 
+    def __repr__(self):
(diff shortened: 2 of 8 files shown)


examples/reachable.myl
```diff
@@ -9,6 +9,6 @@ DO
                               EMIT Edge.dst AS addr]);
     Delta = DIFF(NewlyReachable, Reachable);
     Reachable = UNIONALL(Delta, Reachable);
-WHILE [*COUNTALL(Delta) > 0];
+WHILE [FROM COUNTALL(Delta) AS size EMIT *size > 0];
```
dhalperi: If I understand the language change correctly, basically you only get the "super-duper-concise" `*x` if `x` is already an alias in the global state. You can't unbox a singleton in the emit clause that you're also defining in the emit clause.

I'm okay with that!

raco/myrial/emitarg.py
```diff
@@ -119,6 +123,9 @@ def __init__(self, relation_name):
     def expand(self, symbols):
         return expand_relation(self.relation_name, symbols)
 
+    def __repr__(self):
+        return 'TableWildcardEmitArg(%s)' % self.relation_name
```
dhalperi: `%s` maybe, for the `repr` here?

raco/myrial/query_tests.py
```diff
@@ -667,11 +667,11 @@ def test_unbox_arbitrary_expression(self):
              x[0] > len(self.dept_table)])
         self.check_result(query, expected)
 
-    def test_unbox_inline_table_literal(self):
+    def test_inline_table_literal(self):
         query = """
         emp = SCAN(%s);
         dept = SCAN(%s);
-        out = [FROM emp WHERE id > *[1,2,3].$2 EMIT emp.id];
+        out = [FROM emp, [1,2,3] as tl WHERE id > tl.$2 EMIT emp.id];
```
dhalperi: did you mean to drop the `*`? It looks like the grammar would still be legal.

examples/reachable.myl
dhalperi: I almost suggested we disallow the `*` if `x` is already in the `from_args`, but that probably breaks the case of multiple invocations of `*x.avg, *x.std` .. right?

7andrew7: Yes, that's correct.  Basically, the star operator just does an implicit addition to the FROM clause, assuming the relation hasn't already been imported.  The star is idempotent for the reasons you describe.

raco/myrial/query_tests.py
7andrew7: Yes, this would work either way.

PR #387, opened 2015-02-09
Myria insert split

dhalperi:
@jingjingwang @stechu I think we've talked about these changes.

When a fragment contains multiple heavyweight operators (e.g., join followed by join), we can insert LocalMultiwayProducer to split them into two threads on the same machine. This PR implements this change.

### DIFF
--- a/raco/algebra.py
+++ b/raco/algebra.py
@@ -1087,6 +1087,18 @@ class Broadcast(UnaryOperator):
         return self.opname()
 
 
+class Split(UnaryOperator):
+    """An in-memory pipeline between two operators. Typically used in
+    multi-threaded systems for IPC between threads executing different
+    operator subtrees."""
+
+    def num_tuples(self):
+        return self.input.num_tuples()
+
+    def shortStr(self):
+        return self.opname()
+
+
 class PartitionBy(UnaryOperator):
 
     """Send input to a server indicated by a hash of specified columns."""
--- a/raco/fakedb.py
+++ b/raco/fakedb.py
@@ -367,6 +367,12 @@ class FakeDatabase(Catalog):
     def myriahypershuffleproducer(self, op):
         return self.evaluate(op.input)
 
+    def myriasplitconsumer(self, op):
+        return self.evaluate(op.input)
+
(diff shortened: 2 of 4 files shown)


dhalperi:
See https://demo.myria.cs.washington.edu/profile?queryId=62190# (old way) vs http://localhost:8089/profile?queryId=62208 (new way). The latter finishes much faster because the `Store` and the `LeapFrogJoin` are in different threads... not to mention the sorts happen in different threads as well.

dhalperi:
For now, triggers a bug in myria. Holding on deploying this permanently.

stechu:
:+1: 
Also, have you considered one to many split? Like local broadcast.

dhalperi:
@stechu can you give an example where this should happen now?

stechu:
I have tried to hardcode the parallel semijoin reduction algorithm in [GYM](http://arxiv.org/abs/1410.4156).

For example, if we have query

```datalog
Q(x, y, z, p) :- A(x, y), B(y, z), C(z, p)
```

A generalized hypertree decomposition of the query is:
..............................B(y, z)
................................ / .....\
........................A(x, y)..... C(z, p)

Then the join steps are:

1. Bottom-up semi joins:  
    ```datalog
    B'(y, z) :- A(y), B(y, z)
    B''(y, z) :- C(z), B''(y, z)
    ```

2. Top-down semi joins
    ```datalog
    A'(x, y) :- B''(y), A(x, y)
    C'(z, p) :- B''(z), C(z, p)
    ```

3. Do the actual joins on reduced relations. e.g. `A', B'', C'`

In step 2. we actually need to split B, since B is the source of two different braches of dataflow afterward.

dhalperi:
@stechu that example makes sense -- but I think that's "common subexpression elimination", which is an entirely different optimization than what I'm going for here.

(Otherwise, you could just compute B twice, right? This is what we do now in many cases.)

stechu:
@dhalperi 
Exactly, that is what I do now.
"common subexpression elimination" is indeed more complicated since there are pullings from more than one operators. We need to have in memory data store to enable that. This is definitely beyond this PR.

PR #394, opened 2015-03-10
Code/test changes to enable modulo operator.

senderista:
Not sure whether I did the right thing at https://github.com/uwescience/raco/compare/mod_op#diff-e9797c7b402b3bc30b2713606bcf6097R160. I thought the Python syntax for the intdiv operator was also part of MyriaL syntax, so the visitor should be using the Python intdiv syntax, not the general div symbol.

I have been unable to get nosetests working on my dev machine despite everything seeming to check out (the version of nose that nosetests expects is installed, I can import it in a python shell and check the version, etc.). I'll need some help to fix this, at least until I get my new laptop.

### DIFF
--- a/c_test_environment/platform_tests.py
+++ b/c_test_environment/platform_tests.py
@@ -332,6 +332,10 @@ class MyriaLPlatformTests(object):
         STORE(out, OUTPUT);
         """, "apply")
 
+    def test_idivide(self):
+        q = self.myrial_from_sql(['T2'], "idivide").replace("/", "//")
+        self.check(q, "idivide")
+
     def test_apply_and_self_join(self):
         q = self.myrial_from_sql(['T3'], "apply_and_self_join")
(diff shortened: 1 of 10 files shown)


dhalperi:
Looks reasonable, but yeah we need a test or two. I'm happy to help get your environment set up.

Are you around? I dropped by the DB lab but did not see you. I am in the Data Science Studio for the rest of today.

dhalperi:
@bmyerz see the change to `CompileExpressionVisitor` that Tobin refers to in the issue description. Do you know what the right behavior is? If you use this to generate C code them I'm not sure you'll get the right thing -- surely `//` is not valid C syntax (it's a comment ;).

dhalperi:
@senderista reminder that all this PR is missing is a test or three

bmyerz:
@dhalperi @senderista The change looks fine. I'll add a change below to make sure Clang doesn't break

PR #395, opened 2015-03-10
Fix boolean type check

dhalperi:
Because of a bug in the type checking, we could accidentally traverse a binary tree in
exponential time instead of linear. The bug was checking the type of the same side
twice instead of checking both sides.

Add a test and fix the bug.


stechu:
:+1:

stechu:
Can a test query containing long enough expr be added so that any exponential time traversal will fail the test?

PR #399, opened 2015-03-23
myrialang: add limit operator

dhalperi:
and make sure it is tested

@domoritz you mention this most frequently, so I'm assigning it to you :p

Note that such programs are not optimized very well, but `limit` is now implemented. See also [the terrible MyriaX implementation of limit](https://github.com/uwescience/myria/pull/649) attributable to push-based networking.

### DIFF
--- a/raco/fakedb.py
+++ b/raco/fakedb.py
@@ -360,6 +360,9 @@ class FakeDatabase(Catalog):
     def myriascantemp(self, op):
         return self.scantemp(op)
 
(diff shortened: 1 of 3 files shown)


dhalperi:
@BrandonHaynes the `raco` side of this is here.

domoritz:
A distributed limit would be nice bust I admit that's a bit harder since some nodes may not have enough tuples.

PR #404, opened 2015-04-06
Translate MyriaL programs over RDF to SPARQL backends

### LINKED ISSUE #401: Adding a new backend requires subclassing both Language and CompileExpressionVisitor
If I want to write a custom syntax for a new backend language, I will always need to subclass Language and override body, compile_attribute, etc. 

But if I need to change the syntax for EQ, AND, etc., I appear to need to subclass CompileExpressionVisitor, override the appropriate methods, then override compile_expression in my Language subclass to use the new visitor.

The goal of the Language class was to provide a single point of control for implementing new backend syntaxes.  

So I think CompileExpressionVisitor should call an appropriate Language method in every visit method.

### LINKED ISSUE #403: MyriaL: String literals cannot contain double quotes?
Looking at raco/myrial/parser.py and raco/expression/expression.py, it appears there is no escape syntax for double quotes in string literals.

Wish list: It would be nice to have the choice of single or double quotes a la python.

- [x] fix
- [ ] test(s)

billhowe:
First pass 
No aggregation
No optional / left join

To test, run this:
```bash
$ python scripts/myrial -s examples/rdfsimple.myl
$ python scripts/myrial -s examples/sp2bench/q1.myl
```

You can validate the result [in any SPARQL editor](http://dbpedia.org/sparql).

and some side benefits!
fixes #403 
fixes #401


raco/rules.py
```diff
@@ -406,8 +406,11 @@ def descend_tree(op, cond):
     def fire(self, op):
         if not isinstance(op, algebra.Select):
             return op
-        if op.has_been_pushed:
-            return op
+        if hasattr(op, "has_been_pushed"):
+            if op.has_been_pushed:
+                return op
+            else:
+                 op.has_been_pushed = False
```
bmyerz: This seems redundant. If has_been_pushed is False, make it False. Or is it that has_been_pushed is 0 or None somewhere?

raco/language/__init__.py
```diff
@@ -17,6 +17,19 @@ def opt_rules(self, **kwargs):
...
+    MINUS = "-"
+    DIVIDE = "/"
+    TIMES = "*"
+    IDIVIDE = "//"
+    GT = ">"
+    LT = "<"
+    GTEQ = ">="
+    LTEQ = "<="
+    MOD = "%"
+
```
bmyerz: We already have default literals here https://github.com/uwescience/raco/blob/7c79f8f9adfb1b301106d4932748adb342678281/raco/expression/boolean.py#L52

Perhaps we should use those below instead of self.language.EQ, etc?

raco/rules.py
billhowe: Indeed, I screwed up the indentation when fixing the travis formatting
checks.

The else: should be associated with the hasattr check: If the attribute
isn't present, add it and make it false.

On Mon, Apr 6, 2015 at 9:27 AM, Brandon Myers <notifications@github.com>
wrote:

> In raco/rules.py
> <https://github.com/uwescience/raco/pull/404#discussion_r27808050>:
>
> > @@ -406,8 +406,11 @@ def descend_tree(op, cond):
> >      def fire(self, op):
> >          if not isinstance(op, algebra.Select):
> >              return op
> > -        if op.has_been_pushed:
> > -            return op
> > +        if hasattr(op, "has_been_pushed"):
> > +            if op.has_been_pushed:
> > +                return op
> > +            else:
> > +                 op.has_been_pushed = False
>
> This seems redundant. If has_been_pushed is False, make it False. Or is it
> that has_been_pushed is 0 or None somewhere?
>
> —
> <https://github.com/uwescience/raco/pull/404/files#r27808050>.
>



-- 
Bill Howe
University of Washington
eScience Institute"

raco/language/sparql.py
```diff
@@ -0,0 +1,146 @@
...
+      %s 
+      }
+      """ % (self.left.compile(), self.right.compile())
+
+class SPARQLScan(algebra.Scan, SPARQLOperator):
+    def renameattrs(self):
+        """Make attribute names globally unique so they can be used as SPARQL variables"""
+        global counter
+        c = counter
+        counter = counter + 1
```
bmyerz: Encapsulate this in a gensym function

bmyerz:
Any kind of test we can add to this right now?

scripts/myrial
```diff
@@ -86,7 +90,7 @@ def main(args):
     if os.path.exists(catalog_path):
         catalog = FromFileCatalog.load_from_file(catalog_path)
     else:
-        catalog = FromFileCatalog({})
+        catalog = FromFileCatalog({},"")
```
bmyerz: Looks like FromFileCatalog.__init__ has 1 argument not two. What is this for?

bmyerz:
q2 and q3 are not valid MyriaL syntax

PR #408, opened 2015-04-15
Natively support strings in clang

bmyerz:
Previously strings are supported with an out-of-band indexing step that replaces the strings with integers. This branch adds the string type.

Important:

this branch breaks grappalang: a tuple stores a `STRING_TYPE` as a C++ `std::string`, which has a pointer to a string. In grappalang, since this is not a GlobalAddress, strings won't work.
suggested fix is to replace `std::string` with `std::array<char, N>` so that the data is stored in the Tuple struct

- [x] array refactoring
- [x] test grappa running
- [x] grappa ci tests pass (52)

replaces #400 to change base branch

### DIFF
--- a/c_test_environment/generate_test_relations.py
+++ b/c_test_environment/generate_test_relations.py
@@ -50,15 +50,20 @@ def generate_strings(basename, fields, tuples, datarange):
     fn = get_name(basename, fields)
     with open(fn, 'w') as f:
         print "generating %s" % (os.path.abspath(fn))
+        known_str = "coffee"
+        tuple_with_known = random.randint(0, tuples-1)
         for i in range(0, tuples):
             f.write(str(random.randint(0, datarange)))
             if 0 < (fields - 1):
                 f.write(' ')
             for j in range(1, fields):
-                strmax = 24
-                strlength = random.randint(1, strmax)
-                s = ''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(strlength))
-                f.write(s)
+                if i == tuple_with_known and j == 1:
+                    f.write(known_str)
+                else:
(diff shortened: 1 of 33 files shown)


bmyerz:
String support in grappalang will be more complicated. This is because `std::string` is an object containing heap pointers. 
A few options:
- send `std::string` in a Grappa RPC, and the internal `char*` is turned into a `GlobalAddress` and a second message retrieves it.
- serialize tuples before sending
 - this will complicate the code because we cannot rely on a lambda capture of the tuple or the string
 - we can serialize to a `std::array<SIZE>`
- use `std::array` in the tuple structs
 - this might actually work quite well as string operations are not really needed

billhowe:
On Tue, Apr 14, 2015 at 10:27 PM, Brandon Myers <notifications@github.com>
wrote:

> String support in grappalang will be more complicated. This is because
> std::string is an object containing heap pointers.
> A few options:
>
>    - send std::string in a Grappa RPC, and the internal char* is turned
>    into a GlobalAddress and a second message retrieves it.
>    - serialize tuples before sending
>       - this will complicate the code because we cannot rely on a lambda
>       capture of the tuple or the string
>       - we can serialize to a std::array<SIZE>
>     - use std::array in the tuple structs
>       - this might actually work quite well as string operations are not
>       really needed
>
>
I'm too far away from the code to have a meaningful opinion, but I think
std:array seems reasonable.  If we need string ops down the road, casting
from array to string in the context of the apply op that invokes the string
functions seems feasible.


>
>    -
>
> —
> <https://github.com/uwescience/raco/pull/408#issuecomment-93196706>.
>



-- 
Bill Howe
University of Washington
eScience Institute"

PR #416, opened 2015-05-05
adding ability to print the plan in python repr and read from repr

### LINKED ISSUE #412: Make scripts/myrial more usable for plan manipulations
* --plan will take a python repr as input instead of myrial

bmyerz:
fixes #412 

Note that this feature should only be used in the developer client because it relies on `eval`. It is only exposed in scripts/myrial, which is not used by any server code (myria-web uses raco `processor` API directly)

### DIFF
--- a/c_test_environment/repr_myrial_tests.py
+++ b/c_test_environment/repr_myrial_tests.py
@@ -1,0 +1,27 @@
+import unittest
+from generate_test_relations import generate_default
+from generate_test_relations import need_generate
+from raco.language.myrialang import MyriaLeftDeepTreeAlgebra
+from platform_tests import MyriaLPlatformTestHarness, MyriaLPlatformTests
+from raco.from_repr import plan_from_repr
+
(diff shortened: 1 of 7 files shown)


domoritz:
Thanks for adding the warning!

domoritz:
Do we have tests that make sure that repr actually works for all operators?

bmyerz:
Doing it for all operators _automatically_ is hard even using `Operator.__subclasses__` because different operators take different arguments. I think a reasonable solution is to use the existing test queries and compare the eval(repr(x)) == x

bmyerz:
I don't know what's going on with travis...its running pep on a file that shouldn't exist

senderista:
I couldn't figure that out either, so I just added "--exclude=parsetab.py"
to the pyflakes command:

https://github.com/uwescience/raco/commit/4ac32395018e3159d8b9e12cdc4fa91295b70dab

PR #417, opened 2015-05-05
add the ability to disable optimization

### LINKED ISSUE #414: Make it easy to turn optimizations on and off
Perhaps `Rule`s can automatically generate flags for turning off the optimization.

bmyerz:
fixes #414

### DIFF
--- a/raco/rules.py
+++ b/raco/rules.py
@@ -1,3 +1,5 @@
+import re
+
 from raco import algebra
 from raco import expression
 from .expression import (accessed_columns, UnnamedAttributeRef,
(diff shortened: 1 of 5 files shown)


raco/rules.py
```diff
@@ -15,8 +17,24 @@ class Rule(object):
 
     __metaclass__ = ABCMeta
 
+    _disabled_rules = set()
+    _flag_pattern = re.compile(r'no_([A-Za-z_]+)')  # e.g., no_MergeSelects
+
     def __call__(self, expr):
-        return self.fire(expr)
+        if not hasattr(expr, '_disabled'):
```
domoritz: I'm really not a fan of all this magic. Can you at least write a little comment that explains what is happening and why?

bmyerz: This part is probably an example of premature optimization, so I simplified it below.

PR #420, opened 2015-05-11
Load options

senderista:
@BrandonHaynes I'd like to merge this fix ASAP (before 5/12 if possible).

MyriaX tests for new `load()` syntax to follow shortly.

### DIFF
--- a/examples/load_options.csv
+++ b/examples/load_options.csv
@@ -1,0 +1,4 @@
+this file uses the pipe character ("|") as field delimiter and the tilde ("~") as the quote character
+the percent character ("%") is used to escape the field delimiter
+1|foo|~abc|def~|1.0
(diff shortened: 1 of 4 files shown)


senderista:
I tested the query from `raco/raco/myrial/filescan_tests.py` on `myria-web` and it works fine. Given that `FileScan` already has fairly comprehensive test coverage of its various options, I think more tests in MyriaX are unnecessary. Let me know if you disagree.

```
x = load("file:///Users/tdbaker/raco/examples/load_options.csv",
csv(
    schema(column0:int, column1:string, column2:string, column3:float),
    delimiter="|", quote="~", escape="%", skip=2));
store(x, OUTPUT);
```

BrandonHaynes:
This looks just fine to me!  Merge away.

PR #426, opened 2015-05-20
CLI for uploading data to Radish

bmyerz:
- [x] CLI


domoritz:
Should this be part of raco or live in something like myria-python?

bmyerz:
@domoritz  @billhowe My understanding of myria-python in its current form is that it is talking to myria-web. This CLI talks directly to Raco. In the future, it may be desirable to have myria-python talk to _both_ Raco and myria-web. At that point this CLI can go in myria-python.

billhowe:
This seems right.  Raco for now, but if myria-python becomes our pure local
interface as well (which it may, given all of Brandon Haynes' client-side
work), we can move CLI tools there.

Does that work, Dom?

domoritz:
Sounds reasonable.

PR #431, opened 2015-06-07
Write to Catalog File

jortiz16:
We want to have a way for users to write the schema directly to a file from a call instead of writing it on the file separately. For example: 

```FromFileCatalog.scheme_write_to_file("schema2.py","public:adhoc:newTable", "{'columnNames': ['grpID'], 'columnTypes': ['LONG_TYPE']}")```

```catalog = FromFileCatalog.load_from_file("schema2.py")```

In the future, once we do have a Myria connection in Raco, it would be nice to do something like the following (I'm using Myria-Python high-level calls in this example): 

```connection = MyriaConnection(hostname = "rest.myria.cs.washington.edu", port=1776, ssl=True ...)```

```current_schema = (MyriaRelation(relation="public:adhoc:newTable", connection=connection).schema.to_dict())```

```FromFileCatalog.scheme_write_to_file("schema2.py","public:adhoc:newTable", current_schema)```

### DIFF
--- a/raco/catalog.py
+++ b/raco/catalog.py
@@ -6,6 +6,7 @@ from ast import literal_eval
 import os
 import json
 
+
 class Relation(object):
 
     def __init__(self, name, sch):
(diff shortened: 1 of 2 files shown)


bmyerz:
Can you add a test for
`"{'columnNames': ['grpID'], 'columnTypes': ['LONG_TYPE']}` to file to catalog?

See raco/test_style.py  StyleTest for an example of raco tests

jortiz16:
Sure! I'll get back to this once I wrap up the other PR with delete feature in Myria

PR #440, opened 2015-06-22
fix array equality so that it uses string equality

### LINKED ISSUE #438: Radish string equality check is wrong
Equality is relied on in selections and in hash insertions. We do not want to rely on `std::array::operator==` naively because we really want string equality semantics (that is, the char buffer need only be equal up to the null terminator.

bmyerz:
fixes #438

### DIFF
--- a/c_test_environment/strings.h
+++ b/c_test_environment/strings.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cstring>
 #include <string>
 #include <map>
 #include <vector>
@@ -47,7 +48,23 @@ std::array<char, N> to_array(const Iterable& x) {
   std::array<char, N> d;
   std::copy(x.begin(), x.end(), d.data());
   *(d.data()+x.size()) = '\0'; // copy null terminator
+  
+  // ensure normalization of std::arrays that are equal strings
+  if (x.size()+1 < N) {
+    std::memset(d.data() + x.size() + 1, 0, N - x.size());
+  }
+
   return d;
+}
+
+// utility to see full content of char array 
+template <size_t N>
+void dumps(std::ostream& o, const std::array<char, N>& arr) {
+  o << "[";
+  for (int i=0; i<N; i++) {
+    o << arr[i] << "|";
+  }
+  o << "]";
 }
 
 template <size_t N>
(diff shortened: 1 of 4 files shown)


bmyerz:
difficulty is that Radish hash keys are `std::tuple` and since `std::array<char, N>::operator==` is overriding the default `operator==`, `std::tuple<...>::operator==` does not seem to call the overridden version.

```c
int main() {
  std::array<char, 6> x = {'a','b','c','\0','x','y'};
  std::array<char, 6> y = {'a','b','c','\0','1','2'};
  std::array<char, 6> z = {'a','b','c','\0','x','y'};

  auto xx = std::make_tuple(x);
  auto yy = std::make_tuple(y);

  int status = 0;
  std::cout << abi::__cxa_demangle(typeid(xx).name(), 0, 0, &status) << std::endl;
  std::cout << abi::__cxa_demangle(typeid(yy).name(), 0, 0, &status) << std::endl;

  std::cout << (x==y) << std::endl;
  std::cout << (x==z) << std::endl;


  std::cout << (xx == yy) << std::endl;
```

bmyerz:
works if we also override `std::tuple<...>::operator==`

```c
// necessary to force std::tuple to use overriden std::array
bool operator==(const std::tuple<std::array<char,6>>& a, const std::tuple<std::array<char, 6>>& b) {
  return std::get<0>(a) == std::get<0>(b);
}
```

bmyerz:
The problem is likely that `std::tuple<...>` (and thus std::tuple<...>::operator==) is instantiated before we even have included `"strings.h"`, so the version of `std::tuple<...>` that is instantiated (and thus used by the hash table) doesn't have our `operator==`.

bmyerz:
Other solutions:
* make sure that `std::array` that need strings semantics also are padded with zeroes
* make my own tuples class, so I am defining not overriding `operator==`
* use a different `Pred` for `unordered_map` that usually calls into `std::equal_to` but specialize it for tuple

bmyerz:
checking results
```bash
grep Material tfidf.log | awk '{print $6,$7}' | cut -d'(' -f2 | cut -d',' -f1 | sort | uniq -c | sort | head
```

PR #445, opened 2015-06-24
fields size will include padding between struct elements, just not at…

### LINKED ISSUE #444: fieldsSize is inaccurate because there could be padding between types

bmyerz:
fixes #444


bmyerz:
and also https://github.com/uwsampa/grappa/blob/bmyerz/support-raco-udas/applications/join/Relation_io_tests.cpp#L179

PR #447, opened 2015-06-29
add a way to pass kwargs to the Raco compiler

bmyerz:
example:
```bash
scripts/myrial -c --catalog=examples/catalog.py examples/join.myl --key=scan_array_repr --value=symmetric_array
```


bmyerz:
ok... and a little bit of stuff I forgot from the last PR

PR #450, opened 2015-07-06
Raco-only execution

billhowe:
Restructuring source tree to accommodate connection and catalog objects as well as language objects.

raco/language is now raco/backends, and there are submodules for each backend.

Added raco/backends/myria/connection.py from myria-python and raco/backends/myria/catalog.py from myria-web.  

Added a test in raco/backends/myria/tests/


bmyerz:
the tests are broken; I'll give a shot at fixing

bmyerz:
I don't know that it is common practice to put actual code in `__init__.py` other than import statements. What is the advantage of your approach? I find that it makes some search-based tools harder to use too. I would prefer to use actual names for code. http://www.reddit.com/r/Python/comments/1bbbwk/whats_your_opinion_on_what_to_include_in_init_py/

bmyerz:
@billhowe are there relevant PRs in [myria-web](https://github.com/uwescience/myria-web) and [myria-python](https://github.com/uwescience/myria-python) to reference here?

scripts/simple_raco_execution.py
```diff
@@ -0,0 +1,87 @@
+
+'''
+A notional raco-only script for executing a MyriaL query
```
bmyerz: Is this meant to be executable? I think we should make it so.

Needs `raco.convenience`

raco/backends/myria/tests/test_myria_execution.py
```diff
@@ -0,0 +1,36 @@
+
```
bmyerz: It would be nice to make the connection part of an integration test_suite. I understand that it cannot be a unit test because it depends on a connection to myria.

scripts/simple_raco_execution.py
billhowe: Not yet, but I agree that it still represents the plan. I'd like to get close to an API-compliant refactoring merged first.  We can delete it or rename it for now to avoid confusion?

raco/backends/myria/tests/test_myria_execution.py
billhowe: Added [a test](https://github.com/uwescience/raco/blob/raco_only_execution/raco/backends/myria/tests/test_myria_execution.py) with a connection stub controlled by an environment variable.

scripts/simple_raco_execution.py
billhowe: That's the goal, but I'd like to get this refactoring into master before changing the interface.  (In hindsight, I probably should have put these changes in a separate branch.)  Maybe we can rename this file as a txt to indicate that it is just notional at this point?

bmyerz: I made it print an understandable message if you try to execute

bmyerz:
Looks like the coverage is failing because of new sources with low coverage.

Note the +'s

image

bmyerz:
image

senderista:
fixes #472

senderista:
https://github.com/uwescience/myria/pull/777 is blocking on this PR

bmyerz:
https://coveralls.io/builds/3632714/source?filename=raco%2Fbackends%2Fmyria%2Fconnection.py
https://coveralls.io/builds/3632714/source?filename=raco%2Fbackends%2Fmyria%2Ferrors.py
https://coveralls.io/builds/3632714/source?filename=raco%2Fbackends%2Fmyria%2Fcatalog.py

senderista:
Not sure how to address these coverage errors since all my commit did was
*add* a test...

bmyerz:
these coverage errors originally  came from adding other new files, like the ones I linked to

PR #452, opened 2015-07-07
Changes to run Radish on TPC-H

bmyerz:


bmyerz:
@senderista of all the changes, it is only necessary to review those in:

- raco/backends/backend_common.py
- raco/backends/myria/myria.py
- raco/expression/boolean.py
- raco/expression/expressions_library.py
- raco/expression/function.py
- raco/myrial/parser.py
- raco/myrial/scanner.py
- raco/rules.py
- scripts/myrial

raco/algebra.py
```diff
@@ -1013,6 +1013,13 @@ def scheme(self):
             schema.addAttribute(name, _type)
         return schema
 
+    def __eq__(self, other):
```
senderista: I guess we have to implement `__ne__` as well, right? In the immortal words of the [documentation](https://docs.python.org/2/reference/datamodel.html#object.__ne__): "There are no implied relationships among the comparison operators. The truth of `x==y` does not imply that `x!=y` is false."

Sometimes I think the creator of PHP had more design sense than Guido...

raco/backends/backend_common.py
```diff
@@ -219,6 +231,10 @@ def visit_NaryFunction(self, expr):
                 type(expr).__name__,
                 *arglist))
 
+    def visit_CustomZeroaryOperator(self, expr):
```
senderista: Nitpick: what's the distinction between "function" and "operator" in this context (since the binary form uses "function")?

raco/expression/function.py
```diff
@@ -66,8 +116,19 @@ def typeof(self, scheme, state_scheme):
         return types.DOUBLE_TYPE
 
 
+class YEAR(UnaryFunction):
```
senderista: What's the deal with special-casing this date function? Or is it just a test case for some new feature in this PR?

senderista:
LGTM (I don't understand the `DecomposeGroupBy` code at all, but I guess you just moved it)...

raco/algebra.py
bmyerz: Since we already had `__eq__` before, can we refer this change to #491?

raco/expression/function.py
bmyerz: It was a quick and dirty way to support YEAR, which was needed [in TPC-H queries](https://github.com/uwescience/tpch-radish/blob/7d8e8a8ef5808a6ceccc5d9c4e61b6df7d40f60d/templates/9.sql). I'll add at least the others that make sense.

bmyerz:
RE: DecomposeGroupBy
Yes, the unfortunate thing about the diff is that I both moved and changed that function in the same commit.

```
git diff c1d68de 9100ad4
```

So you cannot see some of the changes I made to it.
If you'd like to you can [download these files](https://gist.github.com/bmyerz/a5819e49d8256d143f40) and use diff.

raco/backends/radish/radish.py
```diff
@@ -1193,6 +1564,17 @@ def opt_rules(self, **kwargs):
         join_type = kwargs.get('join_type', GrappaHashJoin)
         scan_array_repr = kwargs.get('scan_array_repr',
                                      _ARRAY_REPRESENTATION.GLOBAL_ARRAY)
+        groupby_sematics = kwargs.get('groupby_semantics', 'global')
```
senderista: typo

PR #460, opened 2015-07-25
Support for OPP file format in MyriaFileScan.

senderista:
No test coverage since this can't be reasonably implemented in fakedb. Manually tested json query plan generation.


jhyrkas:
LGTM.

PR #467, opened 2015-08-17
proposed design for partitioning info in Raco

### LINKED ISSUE #466: MyriaL rule should extract partitioning info from exchange operators and propagate to Store operator
From @jingjingwang:

> I think a neat way of implementing it might be: adding a rule in raco to extract the partition function from a ShuffleProducer and send it to a Store (and we need to add the PF as a field in DbInsert in MyriaX), then when the server wants to update the metadata of a stored relation, write the PF into catalog. Otherwise, the code may look ugly if we do it purely on the MyriaX side, since we need to dig into query plans.

bmyerz:
Proposing to add interesting orders to the scheme rather than the catalog. The reason is that we would like to track interesting orders at the operator granularity and every operator already has a scheme

__WIP__
- [x] design agreed upon
- [x] @bmyerz make __fake__ catalog create schema with interesting orders
- ~~[ ] @senderista update myria catalog interface to get myria's catalog partition info (note that the myria catalog interface is buried in another unmerged branch https://github.com/uwescience/raco/blob/raco_only_execution/raco/backends/myria/catalog.py part of PR #~450 )~~, moved to #472
- [x] @bmyerz operators propagate partition info from Shuffle/(or radish HashJoin) to store
- ~~[ ] @senderista myriax store `compileme` to json needs to include partitioning info~~, moved to #472
- [x] @bmyerz rule to omit a shuffle when relations are co-partitioned
- [x] tests

We can be conservative, then add more partitioning logic after this PR.

fixes #466

### DIFF
--- a/raco/algebra.py
+++ b/raco/algebra.py
@@ -387,7 +387,7 @@ class IdenticalSchemeBinaryOperator(BinaryOperator):
 
     def partitioning(self):
         """keep the partitioning if both sides are identically partitioned"""
-        lp = self.left.partitioning().hash_partitioned
+        lp = self.left.partitioning()
         if lp.hash_partitioned == self.right.partitioning().hash_partitioned:
             return lp
         else:
--- a/raco/algebra.py
+++ b/raco/algebra.py
@@ -1,6 +1,7 @@
 from raco import expression
 from raco import scheme
 from raco.utility import Printable, real_str
+from catalog import InterestingProperties
 
 from abc import ABCMeta, abstractmethod
 import copy
@@ -72,6 +73,12 @@ class Operator(Printable):
     @abstractmethod
     def num_tuples(self):
         """Return the expected number of tuples output by this operator."""
+
+    @abstractmethod
+    def partitioning(self):
+        """Return the partitioning of the tuples output by this operator.
+        Default implementation returns no information"""
+        return InterestingProperties()
 
     def postorder(self, f):
         """Postorder traversal, applying a function to each operator.  The
@@ -344,6 +351,14 @@ class NaryJoin(NaryOperator):
         # TODO: use AGM bound (P10 in http://arxiv.org/pdf/1310.3314v2.pdf)
         return DEFAULT_CARDINALITY
 
+    def partitioning(self):
+        """attributes are mutually exclusive so union the partitioned
+        attributes"""
+        attributes = set()
+        for c in self.children():
+            attributes.union(c.partitioning().hash_partitioned)
+        return InterestingProperties(hash_partitioned=attributes)
+
     def scheme(self):
         combined = reduce(operator.add, [c.scheme() for c in self.children()])
         # do projection
@@ -366,7 +381,22 @@ class NaryJoin(NaryOperator):
 """Logical Relational Algebra"""
 
 
-class Union(BinaryOperator):
+class SymmetricBinaryOperator(object):
+    """BinaryOperator where both sides have the same schema"""
+
+    def partitioning(self):
+        """attributes that are partitioned in both inputs"""
+        return InterestingProperties(hash_partitioned=set.intersection(
+            self.left.partitioning().hash_partitioned,
+            self.right.partitioning().hash_partitioned))
+
+    def scheme(self):
+        """Same semantics as SQL: Assume first schema "wins" and throw an
+        error if they don't match during evaluation"""
+        return self.left.scheme()
+
+
+class Union(SymmetricBinaryOperator):
 
     """Set union."""
 
@@ -377,16 +407,11 @@ class Union(BinaryOperator):
         # a heuristic
         return int((self.left.num_tuples() + self.right.num_tuples()) / 2)
 
-    def scheme(self):
-        """Same semantics as SQL: Assume first schema "wins" and throw an
-        error if they don't match during evaluation"""
-        return self.left.scheme()
-
     def shortStr(self):
         return self.opname()
 
 
-class UnionAll(BinaryOperator):
+class UnionAll(SymmetricBinaryOperator):
 
     """Bag union."""
 
@@ -400,14 +425,11 @@ class UnionAll(BinaryOperator):
         """deep copy"""
         BinaryOperator.copy(self, other)
 
-    def scheme(self):
-        return self.left.scheme()
-
     def shortStr(self):
         return self.opname()
 
 
-class Intersection(BinaryOperator):
+class Intersection(SymmetricBinaryOperator):
 
     """Set intersection."""
 
@@ -417,14 +439,11 @@ class Intersection(BinaryOperator):
     def num_tuples(self):
         return min(self.left.num_tuples(), self.right.num_tuples())
 
-    def scheme(self):
-        return self.left.scheme()
-
     def shortStr(self):
         return self.opname()
 
 
-class Difference(BinaryOperator):
+class Difference(SymmetricBinaryOperator):
 
     """Set difference"""
 
@@ -435,9 +454,6 @@ class Difference(BinaryOperator):
         left_num = self.left.num_tuples()
         right_num = self.right.num_tuples()
         return left_num - math.floor(min(right_num, left_num * 0.5))
-
-    def scheme(self):
-        return self.left.scheme()
 
     def shortStr(self):
         return self.opname()
@@ -462,6 +478,15 @@ class CompositeBinaryOperator(BinaryOperator):
         return expression.EQ(expression.UnnamedAttributeRef(col0),
                              expression.UnnamedAttributeRef(col1))
 
+    def partitioning(self):
+        """ The schemas are mutually exclusive so union the partition attributes"""
+        return InterestingProperties(hash_partitioned=set.union(self.left.partitioning(),
+                                                                self.right.partition()))
+
+    def scheme(self):
+        """Return the scheme of the result."""
+        return self.left.scheme() + self.right.scheme()
+
 
 class CrossProduct(CompositeBinaryOperator):
 
@@ -479,10 +504,6 @@ class CrossProduct(CompositeBinaryOperator):
 
     def shortStr(self):
         return self.opname()
-
-    def scheme(self):
-        """Return the scheme of the result."""
-        return self.left.scheme() + self.right.scheme()
 
     def add_equijoin_condition(self, col0, col1):
         """Convert the cross-product into a join whenever possible."""
@@ -513,10 +534,6 @@ class Join(CompositeBinaryOperator):
 
     def shortStr(self):
         return "%s(%s)" % (self.opname(), self.condition)
-
-    def scheme(self):
-        """Return the scheme of the result."""
-        return self.left.scheme() + self.right.scheme()
 
     def add_equijoin_condition(self, col0, col1):
         condition = self.get_equijoin_condition(col0, col1)
@@ -598,6 +615,10 @@ class Apply(UnaryOperator):
                      for (name, expr) in self.emitters]
         return scheme.Scheme(new_attrs)
 
+    def partitioning(self):
+        # TODO pass on partitioning for easy cases like renames
+        return InterestingProperties()
+
     def shortStr(self):
         estrs = ",".join(["%s=%s" % (name, str(ex))
                           for name, ex in self.emitters])
@@ -673,6 +694,10 @@ class StatefulApply(UnaryOperator):
     def num_tuples(self):
         return self.input.num_tuples()
 
+    def partitioning(self):
+        # TODO pass on partitioning for easy cases like renames
+        return InterestingProperties()
+
     def copy(self, other):
         """deep copy"""
         self.emitters = other.emitters
@@ -706,6 +731,9 @@ class Distinct(UnaryOperator):
         # TODO: better heuristics?
         return self.input.num_tuples()
 
+    def partitioning(self):
+        return self.input.partitioning()
+
     def scheme(self):
         """scheme of the result"""
         return self.input.scheme()
@@ -730,6 +758,9 @@ class Limit(UnaryOperator):
 
     def num_tuples(self):
         return self.count
+
+    def partitioning(self):
+        return self.input.partitioning()
 
     def copy(self, other):
         self.count = other.count
@@ -778,6 +809,9 @@ class Select(UnaryOperator):
         """scheme of the result."""
         return self.input.scheme()
 
+    def partitioning(self):
+        return self.input.partitioning()
+
     def get_unnamed_condition(self):
         """Get the filter condition for this Select after ensuring that all
         attribute references are UnnamedAttributeRefs."""
@@ -798,6 +832,12 @@ class Project(UnaryOperator):
 
     def num_tuples(self):
         return self.input.num_tuples()
+
+    def partitioning(self):
+        all_columns = set(self.columnlist)
+        return InterestingProperties(hash_partitioned=set.intersection(
+            all_columns,
+            self.input.partitioning().hash_partitioned))
 
     def shortStr(self):
         return "%s(%s)" % (self.opname(), real_str(self.columnlist,
@@ -860,6 +900,13 @@ class GroupBy(UnaryOperator):
         if not self.grouping_list:
             return 1
         return self.input.num_tuples()
+
+    def partitioning(self):
+        all_groupings = set(self.grouping_list)
+        return InterestingProperties(hash_partitioned=set.intersection(
+            all_groupings,
+            self.input.partitioning().hash_partitioned
+        ))
 
     def shortStr(self):
         return "%s(%s; %s)" % (self.opname(),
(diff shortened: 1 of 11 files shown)


senderista:
I vote for something like "RepresentationProperties" since "interesting" is a term of art that won't make sense to those unfamiliar with the historical RDBMS literature.

bmyerz:
From meeting 2015-8-24 
How to provide `get_partitioning`
- logical shuffle can partition by attribute
- most logical operators just always pass the partitioning through (or perhaps always pass on no partitioning)

bmyerz:
(@billhowe 's notes from the myria meeting 2015-8-24)
** Brandon M will add a conservative "partitionedby" method to every logical operator (via Zeroary, Unary, Binary).  Unary will pass through information from its input.  Binary will assume no information (empty list).  Zeroary will assume no information (at first; see below).  Empty list makes sense; from the github issue: "The invariant is that an attribute in the list is known to be partitioned and an attribute not in the list is unknown to be partitioned."  So empty list means "no information."

** For the shuffle operator, partitionedby will return the list of shuffle attributes. (Shuffle is a logical operator currently but probably shouldn't be once we figure out a better way to pass through physical properties from MyriaL to the plan.  The need to explicitly express a specific partitioning appears to have motivated it.  But having "purely logical plans" is useful in places as well.  The technical debt here is not yet so high as to motivate changes.) 

** Tobin will change the rule that generates the physical Store operator to add the partitioning information to the json plan.  This extra information will ensure that every stored relation going forward will have partitioning information included.

** Once that's all tested and in place, we can change the Scan operator's implementation of partitionedby to ask the catalog about partitioning info.  The catalog, in turn, will ask the server.  (We didn't determine if this MyriaX call already exists or not -- that may also need to be implemented.)

** Then, we can add the rule that recognizes when relations are co-partitioned and generates a local join, as well as a rule that pushes everything before the shuffle into a SQL query for postgres (with join being the urgent case).

** To add partitioning info to old data, we will do nothing initially.  The catalog can safely and conservatively return "no information" if the server doesn't find the information.  If we want to address this, we can write a little script to run through all old queries, generate the plan, collect the partitioning information implied, and insert it into the catalog.  But the technical debt here is not to extreme.

raco/algebra.py
```diff
@@ -377,16 +407,11 @@ def num_tuples(self):
         # a heuristic
         return int((self.left.num_tuples() + self.right.num_tuples()) / 2)
 
-    def scheme(self):
-        """Same semantics as SQL: Assume first schema "wins" and throw an
-        error if they don't match during evaluation"""
-        return self.left.scheme()
```
billhowe: We could conceivably throw the error here, right?  Check that left=right?

raco/algebra.py
```diff
@@ -462,6 +478,15 @@ def get_equijoin_condition(col0, col1):
...
                              expression.UnnamedAttributeRef(col1))
 
+    def partitioning(self):
+        """ The schemas are mutually exclusive so union the partition attributes"""
+        return RepresentationProperties(hash_partitioned=set.union(self.left.partitioning(),
+                                                                self.right.partition()))
+
+    def scheme(self):
+        """Return the scheme of the result."""
+        return self.left.scheme() + self.right.scheme()
```
billhowe: I'm not seeing why this is correct.  Why is the scheme not just the left?  If this is Symmetric difference, it's still the left.  Is it like a full outer join or something?

raco/algebra.py
```diff
@@ -1104,6 +1161,9 @@ def num_tuples(self):
     def shortStr(self):
         return self.opname()
 
+    def partitioning(self):
+        return self.input.partitioning()
+
 
 class PartitionBy(UnaryOperator):
```
billhowe: What's the difference between this and Shuffle?

raco/algebra.py
bmyerz: `SymmetricBinaryOperator` takes two identical schemes. Difference falls under this category. Maybe I'll refactor to `SymmetricSchemeBinaryOperator` to avoid confusion

`CompositeBinaryOperator` strictly concats two schemas.

raco/algebra.py
bmyerz: I interpreted Split to be a local operation (no communication cross-worker) used for splitting a fragment in two pieces. Is this not the case?

raco/algebra.py
bmyerz: yes!

raco/algebra.py
billhowe: I see.  The names are ok I think.

raco/algebra.py
billhowe: I see.  I did not read the class carefully.

bmyerz:
above commit also fixes all shuffle rules to be correct. Before they relied on the assumption that a preceding shuffle is the one you wanted.

senderista:
looks like there are a bunch of reproducible test failures

raco/algebra.py
```diff
@@ -344,6 +351,14 @@ def num_tuples(self):
         # TODO: use AGM bound (P10 in http://arxiv.org/pdf/1310.3314v2.pdf)
         return DEFAULT_CARDINALITY
 
+    def partitioning(self):
+        """attributes are mutually exclusive so union the partitioned
+        attributes"""
+        attributes = set()
+        for c in self.children():
+            attributes.union(c.partitioning().hash_partitioned)
```
senderista: How can this work? `set.union(other)` doesn't mutate `self`, it just returns the union of `self` and `other`. Maybe `set.update()` is what you had in mind (I'd still rather avoid mutation for clarity)?

raco/representation.py
```diff
@@ -0,0 +1,17 @@
+__author__ = 'brandon'
+
+
+class RepresentationProperties(object):
+    def __init__(self, hash_partitioned=set(), sorted=None, grouped=None):
+        """
+        @param hash_partitioned: None or list of AttributeRefs in hash key
```
senderista: Isn't this inaccurate given that `hash_partitioned` is an empty set by default and is also a set, not a list?

raco/representation.py
```diff
@@ -0,0 +1,17 @@
+__author__ = 'brandon'
+
+
+class RepresentationProperties(object):
+    def __init__(self, hash_partitioned=set(), sorted=None, grouped=None):
```
senderista: Why is `hash_partitioned` an empty collection by default but the others are not?

raco/algebra.py
```diff
@@ -366,7 +381,29 @@ def shortStr(self):
...
 
 
-class Union(BinaryOperator):
+class IdenticalSchemeBinaryOperator(BinaryOperator):
+
+    """BinaryOperator where both sides have the same schema"""
+
+    def partitioning(self):
+        """attributes that are partitioned in both inputs"""
+        return RepresentationProperties(hash_partitioned=set.intersection(
```
senderista: I don't think the semantics assumed here are correct. Take two relations with identical schema `R1(x, y, z)`, `R2(x, y, z)`, where `R1` is partitioned on `(x, y)` and `R2` is partitioned on `(y, z)`.  This code will make the operator `union(R1, R2)` report that it is partitioned on the attribute `y`, which implies that all tuples in `R1` and `R2` with common values of `y` are present in the same partitions. But that is clearly not the case: a tuple hashed on `(x, y)` will not hash to the same value as a tuple hashed on `(y, z)`. Instead, we can assume identical partitioning if and only if the sets of partitioned attributes in the inputs are identical.

raco/representation.py
BrandonHaynes: Unrelated: initializing `hash_partitioned` with a mutable default argument (evaluated at time of function definition) is probably not what you intended.

raco/algebra.py
```diff
@@ -799,6 +842,12 @@ def __eq__(self, other):
     def num_tuples(self):
         return self.input.num_tuples()
 
+    def partitioning(self):
```
senderista: I don't think this can work for the same reason I stated under `IdenticalSchemeBinaryOperator`.

raco/algebra.py
```diff
@@ -861,6 +910,13 @@ def num_tuples(self):
             return 1
         return self.input.num_tuples()
 
+    def partitioning(self):
```
senderista: I don't think this can work for the same reason I stated under `IdenticalSchemeBinaryOperator`.

raco/algebra.py
```diff
@@ -1092,8 +1171,13 @@ def num_tuples(self):
     def shortStr(self):
         return self.opname()
 
+    def partitioning(self):
```
senderista: Just a random thought: if hash functions were multivalued then maybe they could just emit all workers to simulate broadcast (and a subset of workers for replication and a constant worker for collect)?

raco/algebra.py
senderista: I'm confused--this comment is about `PartitionBy`, not `Split`.

raco/language/myrialang.py
```diff
@@ -547,6 +548,9 @@ def __init__(self, input):
     def num_tuples(self):
         return self.input.num_tuples()
 
+    def partitioning(self):
```
senderista: Shouldn't `partitioning` for the Producer side of an exchange operator reflect the partitioning it will see from its input (i.e., the actual partitioning of tuples on the worker where it's running)?

raco/language/myrialang.py
```diff
@@ -566,6 +570,9 @@ def __init__(self, input):
...
```
senderista: Shouldn't `partitioning` for the Consumer side of an exchange operator reflect the partitioning it imposes on its output (i.e., the actual partitioning of tuples seen on the worker where it runs)?

raco/language/myrialang.py
```diff
@@ -643,6 +656,9 @@ def __repr__(self):
         return "{op}({inp!r}, {hc!r})".format(op=self.opname(), inp=self.input,
                                               hc=self.hash_columns)
 
+    def partitioning(self):
```
senderista: Shouldn't `partitioning` for the Producer side of an exchange operator reflect the partitioning it will see from its input (i.e., the actual partitioning of tuples on the worker where it's running)?

raco/language/myrialang.py
```diff
@@ -684,6 +700,9 @@ def num_tuples(self):
     def shortStr(self):
         return "%s" % self.opname()
 
+    def partitioning(self):
```
senderista: Shouldn't `partitioning` for the Consumer side of an exchange operator reflect the partitioning it imposes on its output (i.e., the actual partitioning of tuples on the worker where it's running)?

raco/language/myrialang.py
```diff
@@ -701,6 +720,10 @@ def __init__(self, input, server):
     def num_tuples(self):
         return self.input.num_tuples()
 
+    def partitioning(self):
```
senderista: Shouldn't `partitioning` for the Producer side of an exchange operator reflect the partitioning it will see from its input (i.e., the actual partitioning of tuples on the worker where it's running)?

raco/language/myrialang.py
```diff
@@ -725,6 +748,9 @@ def __init__(self, input):
...
```
senderista: Shouldn't `partitioning` for the Consumer side of an exchange operator reflect the partitioning it imposes on its output (i.e., the actual partitioning of tuples on the worker where it's running)?

raco/language/myrialang.py
```diff
@@ -754,6 +780,9 @@ def __init__(self, input, hashed_columns,
...
```
senderista: Shouldn't `partitioning` for the Producer side of an exchange operator reflect the partitioning it will see from its input (i.e., the actual partitioning of tuples on the worker where it's running)?

raco/language/myrialang.py
```diff
@@ -781,6 +810,9 @@ def __init__(self, input):
...
```
senderista: Shouldn't `partitioning` for the Consumer side of an exchange operator reflect the partitioning it imposes on its output (i.e., the actual partitioning of tuples on the worker where it's running)?

raco/language/myrialang.py
```diff
@@ -806,6 +838,9 @@ def __repr__(self):
     def num_tuples(self):
         return self._num_tuples
 
+    def partitioning(self):
```
senderista: If, say, a `MyriaScan` (`TableScan` in MyriaX) is optimized to a `MyriaQueryScan` by the `PushIntoSQL` rule, don't we need to preserve the partitioning info from `MyriaScan` for any consumers of the new `MyriaQueryScan`? If so, I guess its constructor should take a `partitioning` arg. Or maybe I'm missing something and all the partitioning info is propagated up the query tree and relevant optimizations are performed before any expressions are optimized into `MyriaQueryScan`.

raco/language/myrialang.py
```diff
@@ -1018,6 +1041,18 @@ def fire(self, exp):
...
 
+def check_partition_equality(op, representation):
+    """Check to see if the operator has the required hash partitioning.
+    @param op operator
+    @param representation list of columns hash partitioned by,
+                        in the unnamed perspective
+    @return true if the op has an equal hash partitioning to representation
+    """
+
+    p = set([expression.ensure_unnamed(attr) for attr in op.partitioning().hash_partitioned])
```
senderista: Nitpick: you can leave out the square brackets in the set constructor, making this a generator expression instead of a list comprehension.

raco/algebra.py
bmyerz: Agreed, I didn't give this much thought; maybe I'll add an example to the comments

raco/language/myrialang.py
bmyerz: I can certainly understand this argument; however I prefer to keep it this way for two reasons:
1. `partitioning()` is a statement about the output of an operator. Either the Consumer or the Producer (or both) must conceptually be responsible for getting partitioned tuples to the destined worker. MyriaX never has anything between a MyriaShuffleProducer and MyriaShuffleConsumer. So, it is valid to say either op starts the partitioned output.
2. So which should we choose? ops like `MyriaShuffleProducer` are the ones in the Producer/Consumer pair that have information about shuffle attributes (notice `MyriaShuffleConsumer` has no members beyond those of `UnaryOperator`), so it is much simpler to allow them to calculate it

raco/language/myrialang.py
bmyerz: see above

raco/language/myrialang.py
bmyerz: see above

raco/language/myrialang.py
bmyerz: see above

raco/language/myrialang.py
bmyerz: see above

raco/language/myrialang.py
bmyerz: see above

raco/language/myrialang.py
bmyerz: see above

raco/language/myrialang.py
bmyerz: I may be mistaken, but I thought `MyriaQueryScan` could push any SQL to Postgres. In general, this could include Apply's and whatever else and so is not trivial like `MyriaScan`

I would propose that if it is important at some point to make this not so conservative that we consider getting the partitioning info from the SQL query fragment's plan
(https://github.com/uwescience/raco/blob/master/raco/language/myrialang.py#L1470).

raco/language/myrialang.py
bmyerz: cool thanks, learned something new about python.

raco/language/myrialang.py
bmyerz: > Or maybe I'm missing something and all the partitioning info is propagated up the query tree and relevant optimizations are performed before any expressions are optimized into MyriaQueryScan.

No, your first interpretation is correct. `partitioning` is functional like `num_tuples`, so what matters is the structure of the AST at the moment you call `MyriaStore.partitioning()` when you are compiling it to json.

raco/language/myrialang.py
senderista: Fair enough, the producer is certainly the only one with the partitioning information.

raco/language/myrialang.py
senderista: I don't quite understand what "getting the partitioning info from the SQL query fragment's plan" means. It looks like the "plan" is just a SQLAlchemy expression, which wouldn't know anything about the partitioning info in the catalog. I'm not suggesting trying to extract partitioning info from `MyriaQueryScan`, which I agree would be intractable, but rather pushing partitioning info into its constructor (https://github.com/uwescience/raco/blob/master/raco/language/myrialang.py#L1473):

```
return MyriaQueryScan(sql=sql_string.process(sql_plan),
                                  scheme=expr.scheme(),
                                  partitioning=expr.partitioning(),
                                  num_tuples=expr.num_tuples())
```

senderista:
Can we add a minimal unit test with this PR that uses the fake catalog and just checks the main use case of eliding shuffles after joins on identically partitioned relations? Coverage for other scenarios could be deferred in a GH issue.

raco/language/myrialang.py
bmyerz: My link above was unclear: the plan I was referring to was `expr` in
```
 sql_plan = cat.get_sql(expr)
```
I've made the change in the latest commit

senderista:
The tests look great and all the checklist items have been completed. Are you OK to merge?

bmyerz:
Thanks for the great feedback @senderista

PR #480, opened 2015-09-23
radish decomposed aggregates

### LINKED ISSUE #478: Radish aggregate combiner step
# Why
In many cases the number of unique keys seen by a given worker may be small enough to fix in local memory. Local combining before communication is a common optimization that takes advantage of these cases.

## Immediate uses
Making TPC-H Q1, Q4 faster (known constant number of keys)
Making matrix multiply queries faster (sometimes sufficiently small number of keys)

# What
Implement GroupBy as localGroupBy=>Shuffle=>localGroupBy in Radish.

# Notes
Notes from email discussion:

The general rule is DecomposeGroupBy and it actually acts on the LogicalAlgebra to turn GroupBy into GroupBy=>Shuffle=>GroupBy, so we still don't explicitly need a different parallel algebra.

The main friction point is just that Radish compiler assumes GroupBy means global groupby, while myria compiler assumes groupby means local groupby. This is where myrialang.py currently abuses the LogicalAlgebra classes by interpreting it in two different ways: before optimization it is a purely logical plan where groupby means full groupby, and after some optimization (translation) steps (fortunately encapsulated in `opt_grps_sequence`) it means partition-local groupby.

All shuffle-y operations in RadishAlgebra are hidden by global operators like RadishHashJoin and RadishGroupby. A pipelined shuffle is easy to implement in Radish (just pull out some code from those data structures that does the RPC call). Then I can introduce two kinds of operators: e.g. RadishGlobalGroupby and RadishPartitionGroupby.

Then optimization can choose either

Groupby =>    RadishPartitionGroupby(Shuffle(RadishPartitionGroupby)))
or
Groupby =>    RadishGlobalGroupby

# To do
- [x] incude myrialang.py `distributed_group_by` rule sequence in `GrappaAlgebra`
- [x] include a way to turn on/off `distributed_group_by` (should be based on cardinality but this is a TODO in DecomposeGroupBy since we don't have good statistics support anyway)
- [x] implement GrappaShuffle; should be just `Grappa::call` where lambda capture is just the input tuple
- [x] Refactor GrappaGroupBy into GrappaGlobalGroupBy
- [x] implement GrappaPartitionGroupBy
- [x] change the translation rule for GroupBy => GrappaGroupBy (must disambiguate partition/global interpretation of GroupBy)

bmyerz:
fixes #478 

Making Radish support decomposed groupbys was relatively simple. It involved 
- implementing `GrappaShuffle` (the code for which is inside of the implementations of Grappa global hash tables).
- pulling `DecomposeGroupBy` rule out of myrialang to rules.py, where it can be used by any distributed algebra
- implementing `GrappaPartitionGroupBy` as a simple override of some code templates from `GrappaGroupBy`

### DIFF
--- a/raco/language/grappa_templates/groupby/withkey_decl.cpp
+++ b/raco/language/grappa_templates/groupby/withkey_decl.cpp
@@ -1,0 +1,2 @@
+decltype(DHT_symmetric<{{keytype}},{{valtype}},hash_tuple::hash<{{keytype}}>>::create_DHT_symmetric( )) {{hashname}};
+
--- a/raco/language/grappa_templates/groupby/withkey_init.cpp
+++ b/raco/language/grappa_templates/groupby/withkey_init.cpp
@@ -1,1 +1,4 @@
-auto {{hashname}} = DHT_symmetric<{{keytype}},{{valtype}},hash_tuple::hash<{{keytype}}>>::create_DHT_symmetric( );
+auto l_{{hashname}} = DHT_symmetric<{{keytype}},{{valtype}},hash_tuple::hash<{{keytype}}>>::create_DHT_symmetric( );
+on_all_cores([=] {
+  {{hashname}} = l_{{hashname}};
+});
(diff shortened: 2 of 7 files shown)


bmyerz:
(apologies for throwing with into bmyerz/tpc-h-fixes...that branch needs to be merged finally)

bmyerz:
result of the optimization on TPC-H

image

PR #481, opened 2015-09-24
Fixes related to rule applications

bmyerz:
- debugging optimizations enhancement
- fix a rule tostring that was reversed and vague
- avoid mutating the tree in rule apply


bmyerz:
minor PR, merging

Issue #484, opened 2015-10-23
union vs. unionall?


jingjingwang:
From https://github.com/uwescience/raco/blob/master/raco/algebra.py#L411 seems `Union` is for sets and `UnionAll` is for bags, which is also how fakedb is implemented: https://github.com/uwescience/raco/blob/master/raco/fakedb.py#L277. But https://github.com/uwescience/raco/blob/master/raco/myrial/parser.py#L800 defines `UnionAll` as a `setop`, and uses "+" as the operator. Then I'm confused... If I write something like `expr_1 + expr_2 + expr_3`, what is the semantics here?

Since `UnionAll` maps to the `UnionAll` operator in MyriaX, I think it should be a `NaryOperator` that unions bags. I'm not sure if we have a use case of `Union`, but if we have, it could be a `BinaryOperator` or a `NaryOperator` that generates a set. On the MyriaX side we can use a `DupElim`, but seems we don't have a mapping for `Union` from raco to MyriaX right now. How do we express 'Union' and 'UnionAll' in MyriaL is another question -- both using "+" sounds not good to me.

billhowe:
It's good to have both UNION and UNIONALL in the algebra, and it's best if
they are binary operators.  MyriaX is not the only backend, and I don't
think most systems consider them as N-ary operators.

The compiler for MyriaX can (and probably should) map a chain of RACO
binary UNIONALLs to a single MyriaX UNIONALL.  That's not a difficult rule
to write.

I'm not a fan of "+" for either UNION or UNIONALL at the syntax level.
Fine to remove it.

"setop" in the parser is just an identifier, isn't it?  We could change it,
but bagop doesn't seem any clearer.

jingjingwang:
I've implemented such a rule to map a chain of UnionAlls to a MyriaX UnionAll, only got confused by Union then. Keeping both set/bag union sounds good, so I guess we just need to agree on the raco syntax. How about using "+" for UnionAll and use the word "Union" for Union (like what we did for other set ops)?

billhowe:
On Fri, Oct 23, 2015 at 12:49 AM, jingjingwang <notifications@github.com>
wrote:

> I've implemented such a rule to map a chain of UnionAlls to a MyriaX
> UnionAll, only got confused by Union then. Keeping both set/bag union
> sounds good, so I guess we just need to agree on the raco syntax. How about
> using "+" for UnionAll and use the word "Union" for Union (like what we did
> for other set ops)?
>

Works for me!

PR #485, opened 2015-10-24
Cleanup union and unionall

jingjingwang:
Based on the discussion in https://github.com/uwescience/raco/issues/484. `Union` is designed for sets and `UnionAll` is designed for bags. We use "+" for `UnionAll` and the word "Union" for `Union`. Clean up all the relevant places, also add an optimization rule to flatten a chain of (binary) `UnionAll`s.


jingjingwang:
@bmyerz I'm not familiar with the grappa/clang side, I tried to find all the places that need to be modified, hopefully they make sense. Let me know if I'm missing anything!

bmyerz:
What is the reasoning for removing `UNIONALL` syntax? This breaks some SQL compatibility. If there is no good reason, can you put in both `+` and `UNIONALL`?

jingjingwang:
Since it's been a while, I rebased my commits on top of master. I added changes to kept both UNIONALL and "+".

raco/algebra.py
```diff
@@ -421,23 +421,37 @@ def shortStr(self):
...
 
     """Bag union."""
 
-    def __init__(self, left=None, right=None):
-        BinaryOperator.__init__(self, left, right)
+    def __init__(self, children=None):
+        NaryOperator.__init__(self, children)
+
+    def partitioning(self):
+        return RepresentationProperties()
```
bmyerz: Is this overly conservative? If all relations are identically partitioned can we keep the partitioning?

raco/backends/cpp/cpp.py
```diff
@@ -548,7 +552,7 @@ def clangify(emit_print):
         rules.OneToOne(algebra.Project, CProject),
         rules.OneToOne(algebra.UnionAll, CUnionAll),
         # TODO: obviously breaks semantics
-        rules.OneToOne(algebra.Union, CUnionAll),
+        rules.OneToOne(algebra.Union, CUnion),
```
bmyerz: I'd prefer to leave Union as not implemented error

jingjingwang:
@bmyerz Done! I made c tests use unionall, also fixed the partitioning of UnionAll. For Datalog tests, I made the IDBs use UnionAll directly, since it's compiled to the physical UnionAll operator anyway.

raco/algebra.py
```diff
@@ -421,23 +421,46 @@ def shortStr(self):
...
+    def partitioning(self):
+        """keep the partitioning if all children are identically
+        partitioned"""
+        for child in self.args:
+            if child.partitioning.hash_partitioned != \
+                    self.args[0].partitioning.hash_partitioned:
+                return RepresentationProperties()
+        return self.args[0].partitioning
+
+    def partitioning(self):
```
bmyerz: ill delete this duplicate

Issue #402, opened 2015-04-06
Add a Left Join operator to Myria algebra?


billhowe:
Unless we have it in some form already.

Sp2Bench is the immediate use case. 

We can fake it with inner join unioned with the stuff that fails to join, but that's a terrifyingly large amount of work compared to just emitting the left tuple padded with nulls instead of dropping it on the floor.

bmyerz:
To be clear, you are proposing to add syntax to MyriaL for a left-join operator and add such an operator to `LogicalAlgebra`?

billhowe:
On Mon, Apr 6, 2015 at 10:40 AM, Brandon Myers <notifications@github.com>
wrote:

> To be clear, you are proposing to add syntax to MyriaL for a left-join
> operator and add such an operator to LogicalAlgebra?
>
I think so, yes.



-- 
Bill Howe
Associate Director and Senior Data Science Fellow, UW eScience Institute
Affiliate Faculty, Computer Science & Engineering
University of Washington
To acknowledge eScience: "Supported in part by the University of Washington
eScience Institute"

domoritz:
Left or right? Myria has a [right hash join](https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/operator/RightHashJoin.java) ;-)

BrandonHaynes:
I also have a use-case where it would be useful to surface a left/right join at the MyriaL level.

billhowe:
Left-only is ok with me.  I find people expect the "root" relation to be on
the left typically.

On Tue, Apr 7, 2015 at 7:14 PM, Dominik Moritz <notifications@github.com>
wrote:

> Left or right? Myria has a right hash join
> <https://github.com/uwescience/myria/blob/master/src/edu/washington/escience/myria/operator/RightHashJoin.java>
> ;-)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/uwescience/raco/issues/402#issuecomment-90781793>.
>



-- 
Bill Howe
Associate Director and Senior Data Science Fellow, UW eScience Institute
Affiliate Faculty, Computer Science & Engineering
University of Washington
To acknowledge eScience: "Supported in part by the University of Washington
eScience Institute"

PR #487, opened 2015-11-01
Move raco docs

### LINKED ISSUE #486: Move raco docs from myria-docs to raco
As we discussed, we want to move component-specific docs into the corresponding repos.

bmyerz:
fixes #486 

retaining history


raco.markdown
```diff
@@ -0,0 +1,89 @@
...
+2. If adding a rule, subclass `Rule` from `raco/rules.py`. You must implemented two methods: `_str_` and `fire`.
+`fire` checks if the rule is applicable to the given tree. If not then it should return the tree itself. If the rule does apply then `fire` should return a transformed tree. It is okay to mutate the input tree and return it: most of Raco's rules are currently doing this instead of keeping the input immutable and copying the whole tree.
+3. Go to your algebra (e.g., `MyriaLeftDeepJoinAlgebra` in `raco/language/myrialang.py`) and instantiate your rule somewhere in the list returned by `opt_rules`.
+
+### Example plan manipulation
+Often users of MyriaX want to partition a table. This can be done by scan the relation, shuffle, store. MyriaL query language does not have a Shuffle operator (they only get introduced by joins, groupbys, and set operations); however we can still easily build the query plan we want. Here is an example of using MyriaX shuffle to partition a table in the MyMergerTree astronomy application.
```
senderista: This is perhaps not the best example since hash partitioning over any set of attributes can be done in `STORE`:
```
T1 = scan(public:vulcan:edgesConnected);
store(T1, public:vulcan:edgesConnectedSort, [$0, $1, $3]);
```

raco.markdown
```diff
@@ -0,0 +1,89 @@
...
+Raco provides many useful rules in `raco/rules.py`. `Rule` is the super class of all rules. 
+
+### How optimization works
+A physical algebra provides an implementation of `opt_rules`, which just returns an ordered list
+of rules to apply. The optimizer applies each rule breadth first to the entire query plan tree, in the order specified by the list.
+This algorithm is very simplistic, but it works out okay right now.
+
+### How to add a rule
+1. first, just check that the rule you need or something very close doesn't already exist in `raco/rules.py` or one of the languages in `raco/language/*.py`. If it is a generic rule and you find it in one of the languages, please [submit a pull request]( moving it to `raco/rules.py`https://github.com/uwescience/raco/compare).
+2. If adding a rule, subclass `Rule` from `raco/rules.py`. You must implemented two methods: `_str_` and `fire`.
```
senderista: "implement"

raco.markdown
```diff
@@ -0,0 +1,89 @@
+# Raco query frontend and middleware
+
+(This document is still a WIP)
+
+[Raco](https://github.com/uwescience/raco) parses query languages, optimizes query plans, and compiles plans to representations accepted by query processing systems.
+
+## Command line usage
+
+`scripts/myrial` provides provides Raco functionality on the command line.
```
senderista: -"provides"

raco.markdown
bmyerz: Do you have any ideas for tasks that are still not covered by rules?

senderista: Well, one case where I had to write a query plan was when I wanted to use a RightHashJoin instead of a SymmetricHashJoin, since I knew one of the relations had much smaller cardinality than the other (and I was actually running out of memory from SymmetricHashJoin persisting all input tuples in hash tables).

bmyerz:
@senderista ready

senderista:
I don't want to bikeshed on this, but I'd like to find a currently-useful example of programmatically munging a query plan, if possible. On the other hand, we need to get this merged. Will you be around for the next Myria student meeting so we could brainstorm a better example? (I gather you don't think that my example of swapping in a different join operator is broadly useful/comprehensible. I would agree that we can't expect non-DB folks to understand/care about the difference between symmetric/asymmetric hash join.)

bmyerz:
There is no MyriaRightHashJoin in Raco for some reason. The other thing I could do with what is already there is replace a shuffle-shuffle symmetric hash join with a broadcast-shuffle symmetric hashjoin

senderista:
There is already an open issue for the missing `MyriaRightHashJoin`: https://github.com/uwescience/raco/issues/402. I just assigned it to myself and plan to submit a PR "soon". I like your idea of replacing shuffle with broadcast in a SymmetricHashJoin: that's something I've done myself (for @sophieclayton) but forgot about, and is genuinely useful. Let's go with that example unless someone has a better one: @billhowe?

raco.markdown
```diff
@@ -0,0 +1,168 @@
...
+import raco.myrial.interpreter as interpreter
+import raco.backends.myria as alg
+from raco.expression.expression import UnnamedAttributeRef
+
+# get the schema
+catalog = FromFileCatalog.load_from_file("vulcan.py")
+_parser = parser.Parser()
+
+# We can have Raco start us with a plan that is close to the one we want by giving it a MyriaL query.
+# Here we start with scan, store. We'll modify it to get scan, shuffle, store.
```
senderista: Just need to tweak this wording a bit (since shuffle is implicit in the join).

bmyerz:
Comments have been addressed, especially the useful example

PR #493, opened 2016-01-19
Fix usage of httmock in myria connection tests

bmyerz:
In travis, the tests using httmock that pass a response body of None, crash. They do not crash in my version of httmock.


bmyerz:
I've recreated the bug on my local machine. It occurs when we upgrade to httmock-1.2.4

PR #503, opened 2016-03-03
Aggregate function coverage push into sql

### LINKED ISSUE #502: `SQLCatalog.get_sql` should support all raco aggregates
We should be able to generate SQL even for aggregates that are not explicitly supported by SQLAlchemy: "Note that any name not known to `func` generates the function name as is - there is no restriction on what SQL functions can be called, known or unknown to SQLAlchemy, built-in or user defined." (http://docs.sqlalchemy.org/en/latest/core/functions.html).

This is a prerequisite for pushing all local `GROUP BY`/`DISTINCT` aggregates into SQL:
https://github.com/uwescience/raco/issues/500
https://github.com/uwescience/raco/pull/501

bmyerz:
fixes #502

### DIFF
--- a/raco/backends/sql/catalog.py
+++ b/raco/backends/sql/catalog.py
@@ -12,6 +12,7 @@ import raco.expression as expression
 import raco.scheme as scheme
 import raco.types as types
 from raco.representation import RepresentationProperties
+import abc
 
 
 type_to_raco = {Integer: types.LONG_TYPE,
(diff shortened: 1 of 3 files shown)


raco/backends/sql/catalog.py
```diff
@@ -107,10 +107,12 @@ def _convert_zeroary_expr(self, cols, expr, input_scheme):
 
     def _convert_unary_expr(self, cols, expr, input_scheme):
         input = self._convert_expr(cols, expr.input, input_scheme)
-        if isinstance(expr, expression.MAX):
-            return func.max(input)
-        if isinstance(expr, expression.MIN):
-            return func.min(input)
+
+        fname = expr.__class__.__name__.lower()
```
senderista: This is certainly an elegant approach, but I think it doesn't work for at least half of the existing aggregates in raco. Specifically, of the unary aggregates in `raco/raco/expression/aggregate.py`, I cannot find a corresponding function for `COUNTALL`, `AVG`, `STDEV` in the list at http://docs.sqlalchemy.org/en/latest/core/functions.html. So I think some special-casing is unfortunately necessary. This is considerably mitigated by the fact that we can just say `func.foo(input)` for any name `foo` and SQLAlchemy will generate the corresponding SQL. So the special cases I think we need to handle right now (for Postgres compatibility) are:
```
AVG -> avg
COUNTALL -> count(*)
STDEV -> stddev_samp
```

bmyerz: Good point. I was coming at this with the backwards assumption that if it isn't in SQLAlchemy then we shouldn't support it. The requirement should be the backend SQL database supports it. For this reason, we should factor out the functionality (ie I'd like PushIntoSQL to be useful for more than just Postgres)

bmyerz: We already take advantage of SQLAlchemy's dialects for compiling a SQLAlchemy expression to a SQL string, but I guess this is orthogonal.

bmyerz: COUNTALL is taken care of already by _convert_zeroary

senderista: Yeah, I don't think we can leverage the dialect support here, unfortunately. How about this: we start with your original approach of testing that the function name is directly supported in SQLAlchemy, and if not, then we special-case by dialect (for now, the only supported dialect is PostgreSQL)? I'm sure you can come up with a nice abstraction for delegating the dialect-driven transformation of the function.

The other question is how to pass the dialect to `SQLCatalog.get_sql()`. I can think of two approaches: we could introduce a new keyword arg `dialect` in `SQLCatalog.__init__()`, or we could add a keyword arg `dialect` to  `SQLCatalog.get_sql()` (defaulting to PostgreSQL in both cases, I guess). What do you think?

senderista: We could also just assume that all the ANSI SQL99 aggregate functions are supported by all SQL backends and translate directly to those without worrying about dialects for now: http://users.atw.hu/sqlnut/sqlnut2-chp-4-sect-2.html. Postgres seems to have pretty good SQL99 conformance (http://www.postgresql.org/docs/9.4/static/features-sql-standard.html), so this may be good enough for now.

raco/backends/sql/catalog.py
```diff
@@ -30,10 +31,44 @@
...
 
 
+class SQLFunctionProvider(object):
+    """Interface for translating function names. For Raco functions
+    not understood by SQLAlchemy, like stddev, we cannot rely
+    on SQLAlchemy's compiler to translate function
+    names to the given dialect.
+    For functions not understood by SQLAlchemy, it just emits them as
+    given."""
+
```
bmyerz: @senderista This comment should explain what I did.

PR #517, opened 2016-05-23
Broadcast support

### LINKED ISSUE #511: Tests for Broadcast
Now that we are [adding broadcast support in the catalog](https://github.com/uwescience/myria/issues/814), I suspect that many tests would break if they were performed on broadcasted input relations. The reason is that Raco doesn't reason in a robust way about broadcasted relations.

bmyerz:
fixes #511

### DIFF
--- a/raco/backends/myria/myria.py
+++ b/raco/backends/myria/myria.py
@@ -154,6 +154,9 @@ class MyriaLanguage(Language):
 class MyriaOperator(object):
     language = MyriaLanguage
 
+    def reduceworkers(self):
+        return None
+
 
 def relation_key_to_json(relation_key):
     return {"userName": relation_key.user,
@@ -169,6 +172,12 @@ class MyriaScan(algebra.Scan, MyriaOperator):
             "relationKey": relation_key_to_json(self.relation_key),
         }
 
+    def reduceworkers(self):
+        if self._debroadcast:
+            return [0]
+        else:
+            None
+
 
 class MyriaScanTemp(algebra.ScanTemp, MyriaOperator):
 
@@ -177,6 +186,12 @@ class MyriaScanTemp(algebra.ScanTemp, MyriaOperator):
             "opType": "TempTableScan",
             "table": self.name,
         }
+
+    def reduceworkers(self):
+        if self._debroadcast:
+            return [0]
+        else:
+            None
 
 
 class MyriaFileScan(algebra.FileScan, MyriaOperator):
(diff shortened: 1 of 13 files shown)


bmyerz:
current state: The functionality should work in many cases. The broadcast plans are not yet thoroughly tested. I'm particularly concerned with the set operators. 

The best tests will be those that show executing the plan gives the correct results. This requires either tests with Myria or enhancing the FakeDB evaluator to consider physical properties.

senderista:
OK, should I pick it up from here (i.e., add tests in MyriaX and/or raco and fix any resulting issues)?

bmyerz:
If we use MyriaX for tests of broadcast, they will have to be system tests, since it involves both repositories. [These examples](https://github.com/uwescience/raco/blob/master/docs/index.md#use-raco-from-python) provide a skeleton of what might go in the test code (using Myria through Raco).

bmyerz:
The failing test on [the last commit](https://github.com/uwescience/raco/pull/517/commits/01a692c869fbc7567b9940e6e306adc990454176)

image

indicates that my DeBroadcast operator is interfering with push Into SQL. It seems like PushIntoSQL rule should go first, following by PushDeBroadcast. PushDeBroadcast will then potentially be pushed into MyriaQueryScan, which enables the worker-0-only optimization.

raco/rules.py
```diff
@@ -1041,3 +1043,55 @@ def check_partition_equality(op, representation):
...
+                copy = MoveDeBroadcast._copy_op(expr.input)
+                copy._debroadcast = True
+                return copy
+            elif isinstance(expr.input, algebra.UnaryOperator):
+                copy = MoveDeBroadcast._copy_op(expr.input)
+                exprcopy = MoveDeBroadcast._copy_op(expr)
+                exprcopy.input = copy.input
+                copy.input = exprcopy
+                return copy
+            elif isinstance(expr.input, algebra.BinaryOperator):
```
senderista: add comment that this case pushes a DeBroadcast op down one level on each branch

raco/backends/myria/myria.py
```diff
@@ -1663,6 +1706,7 @@ def fire(self, expr):
     rules.OneToOne(algebra.Difference, MyriaDifference),
     rules.OneToOne(algebra.OrderBy, MyriaInMemoryOrderBy),
     rules.OneToOne(algebra.Sink, MyriaSink),
+    DeBroadcastToSelect()
```
senderista: Do we actually need this rule given the `MoveDeBroadcast` rule, which pushes `DeBroadcast` ops all the way down to scans, and then sets `overrideWorkers` on those scans to `[0]`? The `DeBroadcastToSelect` rule seems very hackish by comparison.

senderista:
I've addressed my own comments and added enhancements on the `myria_broadcast` branch (which is compatible with the Myria `add_broadcast` branch).

senderista:
superseded by https://github.com/uwescience/raco/pull/548

PR #524, opened 2016-06-06
Add `-D` option to `myrial` for dot of optimized logical plan

dhutchis:


scripts/myrial
```diff
@@ -143,6 +146,12 @@ def main(args):
                 raise "Options dot and --plan are incompatible"
 
             print operator_to_dot(processor.get_logical_plan())
+        elif opt.opt_dot:
+            if opt.from_repr:
```
bmyerz: `from_repr` specifies the *input* not the output. That is, from_repr says take a python string containing a plan rather than a file containing a myrial program.

That is why `--plan` is not in the mutually exclusive arguments group

dhutchis: I don't understand. Are you saying that we should be able to use `myrial -d --plan`?

bmyerz: The way you use plan is as follows:

`myrial --plan file_containing_raco_plan`

instead of

`myrial file_containing_myrial_program`

Therefore it is an argument to specify input and you should be able to use it with any of the output options (the output options are those in the `group`). Maybe the naming of `--plan` is confusing. It means input plan.

dhutchis: Sorry Brandon, I know what you mean by taking in a plan from repr form, as opposed to generating a plan from parsing a MyriaL program.  But I can't get RACO to work with repr's. 

Is this correct?

```bash
myrial -r q9.myl > q9.plan
myrial -d --plan "q9.plan"
```

bmyerz:
Aha, I wrote `--plan` for a very specific purpose: providing a MyriaAlgebra plan so that you can create JSON code.  You can probably leave it as is unless you care about general --plan functionality for more than just MyriaAlgebra. The main purpose of `-r` and `--plan` was to be able to modify plans by hand, but I think people will normally use the python interface rather than the CLI for that.

dhutchis:
Okay, I'm leaving it as is. It would be nice to load any kind of plan from
its repr but I don't want to put in the time to figure out how.

PR #527, opened 2016-07-02
Blob expr udf

parmitam:
Adding support for BLOB type, registering and  retrieving  User defined functions in MyriaX and support for Python UDF expressions.


bmyerz:
I'm not sure whether you are waiting on a code review yet. 
If so, can you fix the tests first?

parmitam:
Brandon, the tests are passing now, sorry about the delay. If you could please take a look, that woudl be great!

PR #531, opened 2016-07-15
RACO support for MyriaX Split operator

senderista:
This is embarrassingly ad-hoc but seems to work end-to-end. I'm not sure it should ever be merged into `master`: instead it should be viewed as a trial run for `FlatMap` support, much like the `Split` operator itself. Note that a debatable design decision I made in the Myria `Split` operator was to add an extra column to the operator output called `X_splits`, where `X` is the name of the column being split. I think I didn't account for this correctly in my (non-)override of `UnaryOperator.scheme()`, but I'm not sure it matters much for now--let me know if you think otherwise.

### DIFF
--- a/raco/algebra.py
+++ b/raco/algebra.py
@@ -618,6 +618,51 @@ def project_partitioning(columnlist, input_partitioning):
         return RepresentationProperties()
 
 
+class StringSplit(UnaryOperator):
+
+    def __init__(self, split_column=None, regex=None, input=None):
+        """Split the specified column of the input using the specified regex.
+
+        :param split_col_index: index of input column to split
+        :type split_col_index: int
+
+        :param regex: valid Java regex
+        :type regex: string
+        """
+
+        UnaryOperator.__init__(self, input)
+        # Raise a TypeError if `splitcol` is not a valid attribute reference in the input schema
+        self.split_column = split_column
+        self.regex = regex
+
+    def __eq__(self, other):
+        return UnaryOperator.__eq__(self, other) and self.split_column == other.split_column and self.regex == other.regex
+
+    def __repr__(self):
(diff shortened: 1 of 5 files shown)


senderista:
Sample query (tested in MyriaWeb with stock MyriaX):

```
R = load("https://s3-us-west-2.amazonaws.com/myria-test/split_test.csv",
csv(schema(a:int, b:int, c:int, d:string),skip=0));
S = split(R, d, ':');
store(S, split_test, [a, b, c]);
```

shrjain:
Just to confirm, this would take any java regex as the third argument right?

senderista:
@shrjain yes, the 3rd argument is a regex with Java syntax

senderista:
Just pushed proper output schema support:

```
R = load("https://s3-us-west-2.amazonaws.com/myria-test/split_test.csv",
csv(schema(a:int, b:int, c:int, d:string),skip=0));
S = split(R, d, ':');
T = [from S emit S.d_splits as s];
store(T, split_test, [s]);
```

shrjain:
Cool. I am hoping to do something like this:
S = split(sample, seq, '/.{11}(.*)/')

Assuming that the kmer length is 11. Very ugly, but I don't have a better way of extracting the first 11 characters for a string.

senderista:
You know we already have `SUBSTR` support, right? Can you explain to me exactly how this kmer extraction is supposed to work? I have barely any idea what it means.

shrjain:
Well, this is embarrassing, I didnt know we had a substr. 
Kmer is just a sliding window substring. 
So for the sequence 'AGCTACAG', 6mers would be:
AGCTAC
GCTACA
CTACAG

senderista:
I'll leave the `split_syntax` branch around, but closing this PR since string split didn't address the use case anyway.

PR #544, opened 2016-12-15
Python UDF support in RACO

BrandonHaynes:
@parmitam : this PR includes a merge of the lambda conversion logic and a number of other commits from `master`, which the new code depends on.  Commit 2f0e8f6bfe1f7fdf5dcdb09c10535a017324e9fa contains all of my changes related to surfacing UDFs in MyriaL.  @shrjain may be in a better position to review the small amount of logic related to lambda conversion, but it is straightforward and unlikely to break (plus there are a bunch of tests).

Note that I changed things up a little bit: instead of invoking a Python UDF as `PYUDF('f', arg0, ..., argn)`, you just invoke it as a function: `f(arg0, ..., argn)`.  This seems like a better design, though `myria-web` will have to be modified to make RACO aware of the available UDFs.


BrandonHaynes:
Made some changes to increase compatibility with `myria-python`, but this PR is still blocking on that project.

parmitam:
Change Bytes_type to BLOB_type, changing the post endpoint to '/function', changing interface to not include inputSchema.

BrandonHaynes:
@parmitam I've modified the Myria backend to support variadic functions, which are resolved at the point of invocation.  This should work for the Python UDFs, which are now implicitly variadic.  I believe that the REST API issues were already corrected.

parmitam:
Looks good. Once we have the myria-python changes made, I'd like to test it end to end with myriaX.

PR #545, opened 2017-01-03
Add recursive idb support

jingjingwang:
This PR adds support for a new syntax: a do-until-convergence loop containing a list of special MyriaL statements. The loop terminates when there will be no update to any relation. Examples are those newly added optimizer tests.

### DIFF
--- a/raco/algebra.py
+++ b/raco/algebra.py
@@ -431,12 +431,12 @@ class IDBController(NaryOperator):
         if expr is None:
             return {"type": "DupElim"}
         if isinstance(expr,
-                      (expression.aggregate.MIN,
-                       expression.aggregate.MULTIMIN)):
+                      (expression.aggregate.MIN, expression.aggregate.LEXMIN)):
             if isinstance(expr, expression.aggregate.MIN):
-                valueCols = [expr.input.position]
+                valueCols = [expr.input.get_position(self.scheme())]
             else:
-                valueCols = [operand.position for operand in expr.operands]
+                valueCols = [operand.get_position(self.scheme())
+                             for operand in expr.operands]
             return {
                 "type": "KeepMinValue",
                 "keyColIndices": group_list,
@@ -452,18 +452,18 @@ class IDBController(NaryOperator):
     def get_group_agg(self):
         agg_list = []
         group_list = []
-        for expr in self.emits:
+        for emit in self.emits:
+            expr = emit.sexprs[0]
             if isinstance(expr, expression.AggregateExpression):
                 if not isinstance(expr, (expression.aggregate.MIN,
-                                         expression.aggregate.MULTIMIN,
+                                         expression.aggregate.LEXMIN,
                                          expression.aggregate.COUNTALL)):
                     raise NotImplementedError(
                         "IDBController does not support agg type {}".format(
                             type(expr)))
                 agg_list.append(expr)
             else:
-                assert isinstance(expr, expression.UnnamedAttributeRef)
-                group_list.append(expr.position)
+                group_list.append(expr.get_position(self.scheme()))
         if len(agg_list) > 1:
             raise NotImplementedError("IDBController only can have one agg")
         if len(agg_list) == 0:
@@ -479,11 +479,27 @@ class IDBController(NaryOperator):
     def scheme(self):
         if ((self.children()[0] is not None) and
                 (not isinstance(self.children()[0], EmptyRelation))):
-            return self.children()[0].scheme()
-        if ((self.children()[1] is not None) and
+            in_scheme = self.children()[0].scheme()
+        elif ((self.children()[1] is not None) and
                 (not isinstance(self.children()[1], EmptyRelation))):
-            return self.children()[1].scheme()
-        return None
+            in_scheme = self.children()[1].scheme()
+        else:
+            return None
+
+        schema = scheme.Scheme()
+        for index, emit in enumerate(self.emits):
+            sexpr = emit.sexprs[0]
+            if isinstance(sexpr, expression.aggregate.LEXMIN):
+                for col in sexpr.operands:
+                    _name, _type = in_scheme.resolve(col)
+                    schema.addAttribute(_name, _type)
+            else:
+                name = (None if emit.column_names is None
+                        else emit.column_names[0])
+                _name = resolve_attribute_name(name, in_scheme, sexpr, index)
+                _type = sexpr.typeof(in_scheme, None);
+                schema.addAttribute(_name, _type)
+        return schema
 
     def shortStr(self):
         return "%s(%s)" % (self.opname(), real_str(self.name,
--- a/raco/algebra.py
+++ b/raco/algebra.py
@@ -70,11 +70,10 @@ class Operator(Printable):
     def walk(self):
         """Return an iterator over the tree of operators."""
         yield self
-        if self.stop_recursion:
-            return
-        for c in self.children():
-            for x in c.walk():
-                yield x
+        if not self.stop_recursion:
+            for c in self.children():
+                for x in c.walk():
+                    yield x
 
     @abstractmethod
     def num_tuples(self):
@@ -89,9 +88,10 @@ class Operator(Printable):
     def postorder(self, f):
         """Postorder traversal, applying a function to each operator.  The
         function returns an iterator"""
-        for c in self.children():
-            for x in c.postorder(f):
-                yield x
+        if not self.stop_recursion:
+            for c in self.children():
+                for x in c.postorder(f):
+                    yield x
         for x in f(self):
             yield x
 
@@ -100,9 +100,10 @@ class Operator(Printable):
         function returns an iterator"""
         for x in f(self):
             yield x
-        for c in self.children():
-            for x in c.preorder(f):
-                yield x
+        if not self.stop_recursion:
+            for c in self.children():
+                for x in c.preorder(f):
+                    yield x
 
     def collectParents(self, parent_map={}):
         """Construct a dict mapping children to parents. Used in
@@ -120,10 +121,9 @@ class Operator(Printable):
         return self.__class__ == other.__class__
 
     def __str__(self):
-        if self.stop_recursion:
-            return self.shortStr()
-        if len(self.children()) > 0:
-            return "%s%s" % (self.shortStr(), real_str(self.children()))
+        if not self.stop_recursion:
+          if len(self.children()) > 0:
+              return "%s%s" % (self.shortStr(), real_str(self.children()))
         return self.shortStr()
 
     def __hash__(self):
@@ -485,12 +485,11 @@ class IDBController(NaryOperator):
         return DEFAULT_CARDINALITY
 
     def scheme(self):
-        if ((self.children()[0] is not None) and
-                (not isinstance(self.children()[0], EmptyRelation))):
-            in_scheme = self.children()[0].scheme()
-        elif ((self.children()[1] is not None) and
-                (not isinstance(self.children()[1], EmptyRelation))):
-            in_scheme = self.children()[1].scheme()
+        child1, child2 = self.children()[:2]
+        if child1 is not None and not isinstance(child1, EmptyRelation):
+            input_scheme = child1.scheme()
+        elif child2 is not None and not isinstance(child2, EmptyRelation):
+            input_scheme = child2.scheme()
         else:
             return None
 
@@ -499,13 +498,14 @@ class IDBController(NaryOperator):
             sexpr = emit.sexprs[0]
             if isinstance(sexpr, expression.aggregate.LEXMIN):
                 for col in sexpr.operands:
-                    _name, _type = in_scheme.resolve(col)
+                    _name, _type = input_scheme.resolve(col)
                     schema.addAttribute(_name, _type)
             else:
                 name = (None if emit.column_names is None
                         else emit.column_names[0])
-                _name = resolve_attribute_name(name, in_scheme, sexpr, index)
-                _type = sexpr.typeof(in_scheme, None)
+                _name = resolve_attribute_name(
+                    name, input_scheme, sexpr, index)
+                _type = sexpr.typeof(input_scheme, None)
                 schema.addAttribute(_name, _type)
         return schema
 
@@ -1252,9 +1252,9 @@ class ProjectingJoin(Join):
     """Logical Projecting Join operator"""
 
     def __init__(self, condition=None, left=None, right=None,
-                 output_columns=None, pull_order=None):
+                 output_columns=None, pull_order_policy='ALTERNATE'):
         self.output_columns = output_columns
-        self.pull_order = pull_order
+        self.pull_order_policy = pull_order_policy
         Join.__init__(self, condition, left, right)
 
     def __eq__(self, other):
@@ -1271,7 +1271,7 @@ class ProjectingJoin(Join):
         return "{op}({cond!r}, {l!r}, {r!r}, {oc!r}, {pr!r})"\
             .format(op=self.opname(), cond=self.condition,
                     l=self.left, r=self.right, oc=self.output_columns,
-                    pr=self.pull_order)
+                    pr=self.pull_order_policy)
 
     def copy(self, other):
         """deep copy"""
@@ -2000,12 +2000,12 @@ class DoWhile(NaryOperator):
 
 class UntilConvergence(NaryOperator):
 
-    def __init__(self, ops=None, pull_order=None):
+    def __init__(self, ops=None, pull_order_policy='ALTERNATE'):
         """Repeatedly execute a sequence of plans until convergence.
         :params ops: A list of operations to execute in parallel.
         """
         NaryOperator.__init__(self, ops)
-        self.pull_order = pull_order
+        self.pull_order_policy = pull_order_policy
 
     def partitioning(self):
         return RepresentationProperties()
--- a/raco/algebra.py
+++ b/raco/algebra.py
@@ -70,6 +70,8 @@ class Operator(Printable):
     def walk(self):
         """Return an iterator over the tree of operators."""
         yield self
+        if self.stop_recursion:
+            return
         for c in self.children():
             for x in c.walk():
                 yield x
@@ -1998,11 +2000,12 @@ class DoWhile(NaryOperator):
 
 class UntilConvergence(NaryOperator):
 
-    def __init__(self, ops=None):
+    def __init__(self, ops=None, pull_order=None):
         """Repeatedly execute a sequence of plans until convergence.
         :params ops: A list of operations to execute in parallel.
         """
         NaryOperator.__init__(self, ops)
+        self.pull_order = pull_order
 
     def partitioning(self):
         return RepresentationProperties()
--- a/raco/algebra.py
+++ b/raco/algebra.py
@@ -206,7 +206,7 @@ class ZeroaryOperator(Operator):
         """Deep copy"""
         Operator.copy(self, other)
 
-    def compileme(self, resultsym):
+    def compileme(self):
         """Compile this operator, storing its result in resultsym"""
         raise NotImplementedError("{op}.compileme".format(op=type(self)))
 
@@ -406,8 +406,9 @@ class ScanIDB(ZeroaryOperator):
         return None
 
     def __repr__(self):
-        return "{op}({name!r},{sch!r})".format(
-            op=self.opname(), name=self.name, sch=self.scheme())
+        return "{op}({name!r},{sch!r},{idb!r})".format(
+            op=self.opname(), name=self.name, sch=self._scheme,
+            idb=self.idbcontroller)
 
 
 class IDBController(NaryOperator):
@@ -424,8 +425,8 @@ class IDBController(NaryOperator):
     def partitioning(self):
         return RepresentationProperties()
 
-    def get_agg_json(self):
-        (group_list, expr) = self.get_group_agg()
+    def get_agg(self):
+        group_list, expr = self.get_group_agg()
         if expr is None:
             return {"type": "DupElim"}
         if isinstance(expr,
@@ -475,11 +476,11 @@ class IDBController(NaryOperator):
         return DEFAULT_CARDINALITY
 
     def scheme(self):
-        if self.children()[0] is not None and not \
-                isinstance(self.children()[0], EmptyRelation):
+        if ((self.children()[0] is not None) and
+                (not isinstance(self.children()[0], EmptyRelation))):
             return self.children()[0].scheme()
-        if self.children()[1] is not None and not \
-                isinstance(self.children()[1], EmptyRelation):
+        if ((self.children()[1] is not None) and
+                (not isinstance(self.children()[1], EmptyRelation))):
             return self.children()[1].scheme()
         return None
 
@@ -518,7 +519,7 @@ class EOSController(UnaryOperator):
         return scheme.Scheme([])
 
     def shortStr(self):
-        return "%s" % (self.opname())
+        return self.opname()
 
 
 class IdenticalSchemeBinaryOperator(BinaryOperator):
@@ -1217,8 +1218,9 @@ class ProjectingJoin(Join):
     """Logical Projecting Join operator"""
 
     def __init__(self, condition=None, left=None, right=None,
-                 output_columns=None):
+                 output_columns=None, pull_order=None):
         self.output_columns = output_columns
+        self.pull_order = pull_order
         Join.__init__(self, condition, left, right)
 
     def __eq__(self, other):
@@ -1232,9 +1234,10 @@ class ProjectingJoin(Join):
                                real_str(self.output_columns, skip_out=True))
 
     def __repr__(self):
-        return "{op}({cond!r}, {l!r}, {r!r}, {oc!r})"\
+        return "{op}({cond!r}, {l!r}, {r!r}, {oc!r}, {pr!r})"\
             .format(op=self.opname(), cond=self.condition,
-                    l=self.left, r=self.right, oc=self.output_columns)
+                    l=self.left, r=self.right, oc=self.output_columns,
+                    pr=self.pull_order)
 
     def copy(self, other):
         """deep copy"""
--- a/raco/algebra.py
+++ b/raco/algebra.py
@@ -381,6 +381,146 @@ class NaryJoin(NaryOperator):
 """Logical Relational Algebra"""
 
 
+class ScanIDB(ZeroaryOperator):
+
+    def __init__(self, name, _scheme=None, idbcontroller=None):
+        self.name = name
+        self._scheme = _scheme
+        self.idbcontroller = idbcontroller
+        ZeroaryOperator.__init__(self)
+
+    def partitioning(self):
+        return RepresentationProperties()
+
+    def num_tuples(self):
+        raise NotImplementedError("{op}.num_tuples".format(op=type(self)))
+
+    def shortStr(self):
+        return "%s(%s)" % (self.opname(), self.name)
+
+    def scheme(self):
+        if self._scheme is not None:
+            return self._scheme
+        if self.idbcontroller is not None:
+            return self.idbcontroller.scheme()
+        return None
+
+    def __repr__(self):
+        return "{op}({name!r},{sch!r})".format(
+            op=self.opname(), name=self.name, sch=self.scheme())
+
+
+class IDBController(NaryOperator):
+
+    def __init__(self, name=None, idb_id=None, children=None, emits=None,
+                 relation_key=None, recursion_mode=None):
+        self.name = name
+        self.idb_id = idb_id
+        self.emits = emits
+        self.relation_key = relation_key
+        self.recursion_mode = recursion_mode
+        NaryOperator.__init__(self, children)
+
+    def partitioning(self):
+        return RepresentationProperties()
+
+    def get_agg_json(self):
+        (group_list, expr) = self.get_group_agg()
+        if expr is None:
+            return {"type": "DupElim"}
+        if isinstance(expr,
+                      (expression.aggregate.MIN,
+                       expression.aggregate.MULTIMIN)):
+            if isinstance(expr, expression.aggregate.MIN):
+                valueCols = [expr.input.position]
+            else:
+                valueCols = [operand.position for operand in expr.operands]
+            return {
+                "type": "KeepMinValue",
+                "keyColIndices": group_list,
+                "valueColIndices": valueCols
+            }
+        if isinstance(expr, expression.aggregate.COUNTALL):
+            return {
+                "type": "CountFilter",
+                "keyColIndices": group_list,
+                "threshold": expr.threshold
+            }
+
+    def get_group_agg(self):
+        agg_list = []
+        group_list = []
+        for expr in self.emits:
+            if isinstance(expr, expression.AggregateExpression):
+                if not isinstance(expr, (expression.aggregate.MIN,
+                                         expression.aggregate.MULTIMIN,
+                                         expression.aggregate.COUNTALL)):
+                    raise NotImplementedError(
+                        "IDBController does not support agg type {}".format(
+                            type(expr)))
+                agg_list.append(expr)
+            else:
+                assert isinstance(expr, expression.UnnamedAttributeRef)
+                group_list.append(expr.position)
+        if len(agg_list) > 1:
+            raise NotImplementedError("IDBController only can have one agg")
+        if len(agg_list) == 0:
+            agg = None
+        else:
+            agg = agg_list[0]
+        return (group_list, agg)
+
+    def num_tuples(self):
+        # TODO
+        return DEFAULT_CARDINALITY
+
+    def scheme(self):
+        if self.children()[0] is not None and not \
+                isinstance(self.children()[0], EmptyRelation):
+            return self.children()[0].scheme()
+        if self.children()[1] is not None and not \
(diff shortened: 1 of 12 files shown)


jingjingwang:
I expect myself adding some more checks to limit what can be put inside of the do-until-convergence loop, but would like to get the main changes being reviewed first.

For 059673c, I'm not sure if it's the best design. I would like to add a MIN aggregate that takes a list of column indices to compare two tuples in that order, however the current MIN is a binary expression. I added a new aggregate called MULTIMIN instead with less support (e.g. fakedb) than MIN, but maybe extending MIN to an nary expression is another option? Would like to hear your thought.

raco/myrial/interpreter.py
```diff
@@ -430,6 +434,114 @@ def dowhile(self, statement_list, termination_ex):
...
 
+    def check_schema(self, expr):
+        return True
+
+    def get_idb_leaves(self, expr, idbs):
+        ret = []
+        op = expr[0].lower()
+        args = expr[1:]
+        if op in ["bagcomp"]:
+            for _id, expr in args[0]:
```
senderista: `expr` is shadowing a function arg

raco/myrial/interpreter.py
```diff
@@ -430,6 +434,114 @@ def dowhile(self, statement_list, termination_ex):
...
+                if expr:
+                    ret += self.get_idb_leaves(expr, idbs)
+                elif _id in idbs:
+                    ret += [_id]
+        elif op in ["select"]:
+            for _id, expr in args[0].from_:
```
senderista: `expr` is shadowing a function arg

raco/myrial/interpreter.py
```diff
@@ -430,6 +434,114 @@ def dowhile(self, statement_list, termination_ex):
...
+
+        done = False
+        while (not done):
+            done = True
+            for _type, _id, emits, expr in statement_list:
+                if idbs[_id].children()[1] is not None:
+                    continue
+                ready = True
+                leaves = self.get_idb_leaves(expr, idbs)
+                for leaf in leaves:
```
senderista: `ready = not any(idbs[leaf].scheme() is None for leaf in leaves)`

raco/myrial/interpreter.py
```diff
@@ -430,6 +434,114 @@ def dowhile(self, statement_list, termination_ex):
...
+                    if idbs[leaf].scheme() is None:
+                        ready = False
+                if ready:
+                    iterative_inputs = self.separate_inputs(expr, idbs, False)
+                    if len(iterative_inputs) == 0:
+                        idbs[_id].children()[1] =\
```
senderista: break line on left paren

raco/myrial/interpreter.py
```diff
@@ -430,6 +434,114 @@ def dowhile(self, statement_list, termination_ex):
...
+                            raco.algebra.EmptyRelation(raco.scheme.Scheme())
+                    elif len(iterative_inputs) == 1:
+                        idbs[_id].children()[1] =\
```
senderista: break line on left paren

raco/myrial/parser.py
```diff
@@ -470,6 +477,12 @@ def p_statement_assign(p):
         p[0] = ('ASSIGN', p[1], p[3])
 
     @staticmethod
+    def p_statement_idbassign(p):
+        'statement : unreserved_id COLON EQUALS LBRACKET emit_arg_list \
```
senderista: Curious why we're introducing a new token `:=` for assignment in this context. Is there some ambiguity if we used `=`? I doubt it since the parse of the `IDBASSIGN` production calls exactly the same code as an ordinary assignment statement. In any case, I find `:=` a little confusing since it's so similar to the Datalog `:-` symbol that we're expressing as `<-`.

raco/algebra.py
```diff
@@ -381,6 +381,146 @@ def shortStr(self):
 """Logical Relational Algebra"""
 
 
+class ScanIDB(ZeroaryOperator):
```
jingjingwang: I think it's because that the plan would have cycles if `IDBController` is the "child", then RACO would apply rules endlessly.

raco/algebra.py
```diff
@@ -104,7 +104,7 @@ def collectParents(self, parent_map=None):
         if parent_map is None:
             parent_map = {}
         for c in self.children():
-            parent_map.setdefault(c, []).append(self)
+            parent_map.setdefault(id(c), []).append(self)
```
jingjingwang: I was following the comment and impl in https://github.com/uwescience/raco/blob/master/raco/backends/myria/myria.py#L1961 (also in function `call_compile_me()` later). I think... yes it looks like a bug to me.

raco/algebra.py
```diff
@@ -381,6 +381,146 @@ def shortStr(self):
...
+        if self.idbcontroller is not None:
+            return self.idbcontroller.scheme()
+        return None
+
+    def __repr__(self):
+        return "{op}({name!r},{sch!r})".format(
+            op=self.opname(), name=self.name, sch=self.scheme())
+
+
+class IDBController(NaryOperator):
```
jingjingwang: Yes.

raco/algebra.py
```diff
@@ -381,6 +381,146 @@ def shortStr(self):
...
+            if isinstance(expr, expression.AggregateExpression):
+                if not isinstance(expr, (expression.aggregate.MIN,
+                                         expression.aggregate.MULTIMIN,
+                                         expression.aggregate.COUNTALL)):
+                    raise NotImplementedError(
+                        "IDBController does not support agg type {}".format(
+                            type(expr)))
+                agg_list.append(expr)
+            else:
+                assert isinstance(expr, expression.UnnamedAttributeRef)
```
jingjingwang: I think it's because that these attributes are indeed unnamed, since the IDBController takes two input relations with same column types but different column names. So we don't know what to call them in a MyriaL statement. Maybe there exists a better solution..?

raco/algebra.py
```diff
@@ -381,6 +381,146 @@ def shortStr(self):
...
+                group_list.append(expr.position)
+        if len(agg_list) > 1:
+            raise NotImplementedError("IDBController only can have one agg")
+        if len(agg_list) == 0:
+            agg = None
+        else:
+            agg = agg_list[0]
```
jingjingwang: Because I want to check if there are more than 1 agg and raise an error if it happens. It's different from normal groupby-agg since we can have more than 1 agg there.

raco/algebra.py
```diff
@@ -1797,6 +1937,28 @@ def scheme(self):
         return None
 
 
+class UntilConvergence(NaryOperator):
```
jingjingwang: Not really empty since `DoWhile` has a check on having >=2 children (the last one is the stopping condition) while `UntilConvergence` does not. I can make them inherit a common class with that additional check if you prefer.

raco/backends/myria/myria.py
```diff
@@ -413,10 +413,40 @@ def compileme(self, leftid, rightid):
...
+
+    def compileme(self, *args):
+        ret = {
+            "opType": "IDBController",
+            "argSelfIdbId": "%s" % self.idb_id,
+            "argInitialInput": "%s" % args[0],
+            "argIterationInput": "%s" % args[1],
+            "argEosControllerInput": "%s" % args[2],
+            "argState": self.get_agg_json(),
+            "sync": self.recursion_mode == "SYNC"
```
jingjingwang: The initial purpose of adding `IDBController` was for async iterations, since `DoWhile` can handle sync already (although it has to store after each iteration). Then we added the sync part to `IDBController` since it's just a small change that fits into the operator design and we think it's convenient to have both in one operator. In terms of worst/best-case performance... we can always make things happen :)

raco/backends/myria/myria.py
```diff
@@ -687,6 +717,10 @@ class MyriaSplitConsumer(algebra.UnaryOperator, MyriaOperator):
     """A Myria SplitConsumer"""
 
     def __init__(self, input):
+        assert isinstance(input, MyriaSplitProducer)
+        if not hasattr(input, 'consumers'):
```
jingjingwang: This is for the rule `RemoveSingleSplit`, which removes a `SplitProducer` that only has one `SplitConsumer` followed by a `ShuffleProducer`. So the `SplitProducer` needs to know its consumers.

raco/backends/myria/myria.py
```diff
@@ -760,11 +794,15 @@ def compileme(self, inputid):
         else:
             raise ValueError("Invalid ShuffleType")
 
-        return {
+        ret = {
             "opType": "ShuffleProducer",
             "argChild": inputid,
             "argPf": pf
         }
+        if hasattr(self, 'buffer_type') and self.buffer_type != "None":
```
jingjingwang: Yep, you can tell from this line that I was a Python beginner at that time :)

raco/backends/myria/myria.py
```diff
@@ -790,6 +828,53 @@ def compileme(self, inputid):
         }
 
 
+class MyriaConsumer(algebra.ZeroaryOperator, MyriaOperator):
```
jingjingwang: This new class is again for breaking the infinite loop of optimization. The query plan being constructed has loops such as: `IDBController` -> (sending EOIs) `Consumer` -> `EOSController` -> (sending EOSs) `Consumer` -> 'IDBController`, so I would need to break the loop to stop the recursive optimization. I ended up with making these consumers as `ZeroaryOperator` to set the optimization boundary. I agree that it's not a neat refactoring but given the current design, I'm not sure if a better solution exists. Maybe we can rename it to be more clear. What do you think?

raco/backends/myria/myria.py
```diff
@@ -1216,6 +1301,22 @@ def __str__(self):
...
 
 
+class ShuffleBeforeIDBController(rules.Rule):
+    def fire(self, expr):
+        # If not a join, who cares?
+        if not isinstance(expr, algebra.IDBController):
+            return expr
+        (group_list, agg) = expr.get_group_agg()
+
+        for idx in range(0, 2):
```
jingjingwang: The 3rd child is a consumer receiving EOS from `EOSController` so it's not a part of this rule (I should just use `range(2)` for sure). For shuffle vs. partial aggregation, it's not always correct to do partial aggregation before `IDBController`, since there are tricky aggs such as `CountFilter`. Another thing is it's not always more efficient to do partial aggregation before global. For example, if all the tuples are unique, then doing a paritial `DupElim` before an `IDBController` with `DupElim` would only double the cost. I think the difference here is that the aggs in `IDBController` and group-by are not the same ones.

I didn't check if eliding the shuffle is possible since I remember that redundant shuffles are removed by some other rules, but if it's not the case (or if it doesn't apply to `IDBController`) then I will add such a check here. (Is it?)

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+
+class ReplaceStoreIDB(rules.Rule):
+
+    @staticmethod
+    def collect_and_replace(op, idbproducers):
+        if not isinstance(op, algebra.NaryOperator):
+            return
+        if isinstance(op, algebra.UntilConvergence):
+            for idbproducer in op.children():
+                if isinstance(idbproducer.input, MyriaIDBController):
```
jingjingwang: It should be a `MyriaSplitProducer` that sits on top of a `IDBController`.

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+        for child in op.children():
+            if isinstance(child, algebra.UntilConvergence):
+                ReplaceStoreIDB.collect_and_replace(child, idbproducers)
+            if isinstance(child, algebra.Store) and \
+                    isinstance(child.input, algebra.ScanIDB):
+                assert child.input.name in idbproducers
+                idbproducers[child.input.name].input.relation_key = \
+                    child.relation_key
+            else:
+                newchildren.append(child)
```
jingjingwang: It could be any operator except `Store`.  We start from the whole MyriaL program and check the children of the current operator. If the child is a `Store` on top of a `ScanIDB`, we discard this store and tell the IDBController to store directly. Otherwise the child, no matter which type it is, is kept.

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+class RemoveSingleSplit(rules.Rule):
+
+    def fire(self, op):
+        if not isinstance(op, algebra.UntilConvergence):
+            return op
+
+        parent_map = {}
+        op.collectParents(parent_map)
+        for idx, child in enumerate(op.children()):
+            if not isinstance(child, MyriaSplitProducer):
```
jingjingwang: `UntilConvergence` has a list of `MyriaSplitProducer` sthat sit on top of `IDBController`s. It is actually also a child of `MyriaConsumer`, but for breaking the infinite optimization loop (see above), I made `MyriaConsumer` as a `ZeroaryOperator` with its producer as a field but as the child (`input`).

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+                continue
+            if len(child.consumers) != 1:
+                continue
+            producer = parent_map[id(child.consumers[0])][0]
+            if not isinstance(producer, MyriaShuffleProducer):
```
jingjingwang: It is, see the newly added `__str__()`.

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+                    expr.input = consumer
+                elif isinstance(expr, algebra.BinaryOperator):
+                    if idx == 0:
+                        expr.left = consumer
+                    else:
+                        expr.right = consumer
+                elif isinstance(expr, algebra.NaryOperator):
+                    expr.args[idx] = consumer
+                if isinstance(expr, algebra.Select) and \
+                        isinstance(expr.condition, expression.boolean.GTEQ):
```
jingjingwang: It is, but the backend can't handle it yet. Since it's on top of `count` which generates integers, `GTEQ` and `GT` should be equivalent. (I need to add it to the backend though)

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+        if not isinstance(op, algebra.UntilConvergence):
+            return op
+        if hasattr(op, "has_been_transformed"):
+            if op.has_been_transformed:
+                return op
+
+        eos_controller = MyriaEOSController()
+        eos_consumers = []
+        idb_controllers = {}
+        for idb_controller in op.args:
```
jingjingwang: `op` is `algebra.UntilConvergence` which has `args`.

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+                        agg.threshold = expr.condition.right.value
+                        expr.condition = None
+            else:
+                DoUntilConvergence.replace_scan_with_idb_consumer(
+                    op, idb_controllers, idb_producers)
+
+    def fire(self, op):
+        if not isinstance(op, algebra.UntilConvergence):
+            return op
+        if hasattr(op, "has_been_transformed"):
```
jingjingwang: I'll remove this flag and check `EOSController` instead.

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+            else:
+                newchildren.append(child)
+        op.args = newchildren
+
+    def fire(self, op):
+        self.collect_and_replace(op, {})
+        return op
+
+
+class RemoveEmptyFilter(rules.Rule):
```
jingjingwang: I'll add one check in `test_galaxy_evolution()`.

raco/backends/myria/myria.py
```diff
@@ -1926,7 +2223,10 @@ def fragments(rootOp):
     def call_compile_me(op):
         "A shortcut to call the operator's compile_me function."
         op_id = op_ids[id(op)]
-        child_op_ids = [op_ids[id(child)] for child in op.children()]
+        if isinstance(op, MyriaConsumer):
```
jingjingwang: See above explanation for `MyriaConsumer`.

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+            return op
+        if hasattr(op, "has_been_transformed"):
+            if op.has_been_transformed:
+                return op
+
+        eos_controller = MyriaEOSController()
+        eos_consumers = []
+        idb_controllers = {}
+        for idb_controller in op.args:
+            idb_controller.children()[2] = MyriaConsumer(None, eos_controller)
```
jingjingwang: I see, should we make it do the right thing?

raco/myrial/parser.py
jingjingwang: The motivation is exactly the same as your concern: I wanted to show that its an IDB :) I think `=` is good and given your previous suggest of adding `idbassign_list` as a new production, parsing should have no problem. I'll change it.

raco/myrial/interpreter.py
jingjingwang: Cool!

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+    @staticmethod
+    def replace_scan_with_idb_consumer(expr, idb_controllers, idb_producers):
+        if isinstance(expr, algebra.ZeroaryOperator):
+            return
+        for idx, op in enumerate(expr.children()):
+            if isinstance(op, algebra.ScanIDB):
+                if op.name not in idb_producers:
+                    idb_producers[op.name] = \
+                        MyriaSplitProducer(idb_controllers[op.name])
+                consumer = MyriaConsumer(None, idb_producers[op.name])
```
jingjingwang: Manually setting up the pair is for breaking infinite optimizations (see above discussion). The rationale is, a `UntilConvergence` consists of a list of `SplitProducer`s, each takes an `IDBController` as the child. The corresponding `SplitConsumer`s are used as the recursive inputs of IDBs (since an IDB may occur in the bodies of many statements).

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
         return algebra.UnionAll(children)
 
 
+class FillInJoinPullOrder(rules.Rule):
```
jingjingwang: I agree it's an awkward design, however exposing it in `DO UNTIL CONVERGENCE` is ugly as well. Ideally this kind of physical level tuning should be done by a cost-based optimizer... On the other hand, it doesn't hurt if we can declare all the joins in a program to pull from some side, it might be a good optimizations too (e.g. saving one hash table when first consuming only one child). So I ended up doing this. I've also thought about not setting this arg in RACO, only doing this manually on JSON. What do you think?

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+    def fire(self, op):
+        if not isinstance(op, MyriaIDBController):
+            return op
+        buffer_type = op.get_agg_json()
+        if buffer_type['type'] == "CountFilter":
+            # do not propagate CountFilter
+            return op
+        # only do it for the iterative input
+        if isinstance(op.args[1], MyriaShuffleConsumer):
+            op.args[1].input.buffer_type = buffer_type
```
jingjingwang: I didn't add `buffer_type` as one arg of `__init__` since this information is not available until `PropagateAsyncFTBuffer`, but `MyriaShuffleProducer`s are constructed before this rule. I added a set method to set the buffer type.

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+                    if idx == 0:
+                        expr.left = child
+                    else:
+                        expr.right = child
+                elif isinstance(expr, algebra.NaryOperator):
+                    expr.args[idx] = child
+        return expr
+
+
+class RemoveSingleSplit(rules.Rule):
```
jingjingwang: I added a check in `test_cc()`.

raco/backends/myria/myria.py
```diff
@@ -1721,6 +1824,193 @@ def fire(self, op):
...
+        eos_consumers = []
+        idb_controllers = {}
+        for idb_controller in op.args:
+            idb_controller.children()[2] = MyriaConsumer(None, eos_controller)
+            eos_consumers.append(MyriaConsumer(None, idb_controller))
+            idb_controllers[idb_controller.name] = idb_controller
+        if len(eos_consumers) == 1:
+            eos_controller.input = eos_consumers[0]
+        else:
+            eos_controller.input = MyriaUnionAll(eos_consumers)
```
jingjingwang: I don't expect `UnionAll` to have only one child in all the other places since it's constructed from `[...] + [...]`, so adding another rule seems unnecessary.

raco/myrial/parser.py
```diff
@@ -409,6 +409,13 @@ def p_optional_arg_list(p):
         p[0] = p[1] or []
 
     @staticmethod
+    def p_recursion_mode(p):
```
jingjingwang: See above discussion.

jingjingwang:
I've addressed most comments except for a few pending discussions:

- The `Zeroary` `MyriaConsumer` for breaking infinite optimization
- `MIN` vs. `MULTIMIN` and should they use named attribute (and what if we don't know the name)
- Join pull order: global or within do-until-convergence only, where to specify

We can discuss them in person if necessary.

raco/expression/aggregate.py
```diff
@@ -203,6 +203,15 @@ def typeof(self, scheme, state_scheme):
         return self.input.typeof(scheme, state_scheme)
 
 
+class MULTIMIN(NaryFunction, TrivialAggregateExpression):
```
jingjingwang: I'm happy to see `MIN` taking multiple columns but sometimes the column names are unknown (see above discussion).

raco/backends/myria/myria.py
```diff
@@ -799,7 +806,7 @@ def compileme(self, inputid):
             "argChild": inputid,
             "argPf": pf
         }
-        if hasattr(self, 'buffer_type') and self.buffer_type != "None":
+        if self.buffer_type is not "None":
```
senderista: You just mean `None`, right?

raco/backends/myria/myria.py
```diff
@@ -1883,23 +1888,33 @@ def fire(self, op):
         self.collect_and_replace(op, {})
         return op
 
+    def __str__(self):
+        return ("Store[relation](IDBScan(IDBController)) ->"
```
senderista: should be trailing space after arrow

raco/backends/myria/myria.py
```diff
@@ -1883,23 +1888,33 @@ def fire(self, op):
...
+                "IDBController[relation]")
+
+
+def replace_child_with(op, child, replacement):
```
senderista: Why don't we `assert` in an `else` clause rather than doing nothing for unexpected operator types.

raco/backends/myria/myria.py
```diff
@@ -1937,51 +1952,50 @@ def fire(self, op):
...
+                         isinstance(op.condition, expression.boolean.GT))):
+                    group_list, agg = idb_controllers[ch.name].get_group_agg()
                     if isinstance(agg, expression.aggregate.COUNTALL):
-                        agg.threshold = expr.condition.right.value
-                        expr.condition = None
+                        if isinstance(op.condition, expression.boolean.GTEQ):
+                            agg.threshold = op.condition.right.value
+                        else:
+                            # count generates integers, translate it to GTEQ
+                            agg.threshold = op.condition.right.value + 1
```
senderista: shouldn't you assert `isinstance(op.condition, expression.boolean.GT)`?

raco/backends/myria/myria.py
```diff
@@ -1937,51 +1952,50 @@ def fire(self, op):
...
-                    op, idb_controllers, idb_producers)
+                    ch, idb_controllers, idb_producers)
 
     def fire(self, op):
         if not isinstance(op, algebra.UntilConvergence):
             return op
-        if hasattr(op, "has_been_transformed"):
-            if op.has_been_transformed:
-                return op
+        if any(isinstance(ch, MyriaEOSController) for ch in op.args):
```
senderista: nitpick: `children()` not `args`

raco/backends/myria/myria.py
```diff
@@ -2003,14 +2017,20 @@ def fire(self, op):
...
         return op
 
     def __str__(self):
         return ("DoUntilConvergence")
 
 
+def idb_until_convergence(**kwargs):
+    ret = [DoUntilConvergence(), RemoveEmptyFilter(), StoreFromIDB(),
+           RemoveSingleSplit()]
+    if kwargs.get('async_ft', 'None') is not None:
```
senderista: Why the second argument? The default of one-arg `get()` is `None`, of course. You could really just say `kwargs.get('async_ft')`, since apparently any non-null value is equivalent.

raco/backends/myria/myria.py
```diff
@@ -2043,23 +2063,20 @@ def opt_rules(self, **kwargs):
...
 
         if kwargs.get('add_splits', True):
             compile_grps_sequence.append([InsertSplit()])
         # Even when false, plans may already include (manually added) Splits,
         # so we always need BreakSplit
         compile_grps_sequence.append([BreakSplit()])
-        compile_grps_sequence.append([RemoveSingleSplit()])
 
-        if kwargs.get('join_pull_order'):
+        if kwargs.get('join_pull_order', 'None') is not None:
```
senderista: Again, the default of `get()` is `None`, so not sure what you're trying to do with the literal string `'None'`.

raco/backends/myria/myria.py
```diff
@@ -2302,4 +2319,4 @@ def compile_to_json(raw_query, logical_plan, physical_plan,
 
     return {"rawQuery": raw_query, "logicalRa": str(logical_plan),
             "language": language, "plan": compile_plan(physical_plan),
-            "ftMode": kwargs.get('async_ft')}
+            "ftMode": kwargs.get('async_ft', None)}
```
senderista: The default of one-arg `get()` is already `None`.

raco/backends/myria/myria.py
jingjingwang: The if condition above ensures that it's either a `GTEQ` or `GT` so I put `else` here.

raco/backends/myria/myria.py
jingjingwang: This was fixed in da94a7f above.

raco/backends/myria/myria.py
jingjingwang: Also fixed in da94a7f.

raco/backends/myria/myria.py
```diff
@@ -1914,14 +1902,44 @@ def fire(self, op):
 
 class FillInJoinPullOrder(rules.Rule):
 
-    def __init__(self, pull_order):
+    def __init__(self):
         self._disabled = False
-        self.pull_order = pull_order
+
+    def get_idb_side(self, op):
```
senderista: This is a misleading name: it doesn't distinguish between "sides" at all, but just tells you if there's an `IDBController` in the current operator's subtree. I think `has_idb_child()` might be a more accurate name.

raco/backends/myria/myria.py
```diff
@@ -1914,14 +1902,44 @@ def fire(self, op):
...
+        if not isinstance(op, algebra.UntilConvergence):
+            return op
+        if op.pull_order is None or op.pull_order == 'ALTER':
             return op
-        op.pull_order = self.pull_order
+        for ch in op.walk():
+            if isinstance(ch, MyriaSymmetricHashJoin):
+                left_idb = self.get_idb_side(ch.left)
+                right_idb = self.get_idb_side(ch.right)
+                if op.pull_order == 'PULL_IDB':
```
senderista: As I mentioned earlier, I think this code would be more readable if `op.pull_order` became `op.pull_order_policy`.

raco/myrial/emitarg.py
```diff
@@ -96,7 +96,7 @@ def get_statemods(self):
         return self.statemods
 
     def __repr__(self):
-        return 'NaryEmitArg(%r)' % self.sexprs
+        return 'NaryEmitArg(%r,%r)' % (self.column_names, self.sexprs)
```
senderista: This is less broken than the original, but still broken. You need to include all constructor args:
```
return 'NaryEmitArg(%r,%r,%r)' % (self.column_names, self.sexprs, self.statemods)
```

raco/myrial/interpreter.py
```diff
@@ -488,18 +488,16 @@ def separate_inputs(self, expr, idbs, is_init):
                 return [expr]
         return []
 
-    def untilconvergence(self, statement_list, recursion_mode):
+    def untilconvergence(self, statement_list, recursion_mode, pull_order):
```
senderista: Per above, `pull_order_policy`.

raco/myrial/optimizer_tests.py
```diff
@@ -1263,22 +1264,27 @@ def test_cc(self):
         """Test Connected Components"""
         query = """
         E = scan(public:adhoc:Z);
-        V = select distinct E.$0 from E;
+        V = select distinct E.src as x from E;
         do
-            CC = [$0, MIN($1)] <- [from V emit V.$0 as x, V.$0 as y] +
-                    [from E, CC where E.$0 = CC.$0 emit E.$1, CC.$1];
-        until convergence;
+            CC = [nid, MIN(cid) as cid] <-
```
senderista: The named attributes are way more readable!

raco/myrial/optimizer_tests.py
```diff
@@ -1325,18 +1331,18 @@ def test_galaxy_evolution(self):
...
+                emit P1.f as time, P1.e as gid1, P2.e as gid2];
+        Galaxies = [time, gid] <-
+          [from GoI emit 1 as time, GoI.a as gid] +
           [from Galaxies, Edges
-           where Galaxies.$0 = Edges.$0 and
-           Galaxies.$1 = Edges.$1 and Edges.$3 >= 4
-           emit Galaxies.$0+1, Edges.$2];
-        until convergence async;
+           where Galaxies.time = Edges.time and
+           Galaxies.gid = Edges.gid1 and Edges.num >= 4
```
senderista: It would be nice to add a test somewhere for the `GT` threshold since we support it as well.

raco/myrial/scanner.py
```diff
@@ -8,7 +8,8 @@
 keywords = ['WHILE', 'DO', 'DEF', 'APPLY', 'CASE', 'WHEN', 'THEN',
             'ELSE', 'END', 'CONST', 'LOAD', 'DUMP', 'CSV', 'SCHEMA',
             'OPP', 'TIPSY', 'UDA', 'TRUE', 'FALSE', 'HASH', 'BROADCAST',
-            'ROUND_ROBIN', 'UNTIL', 'CONVERGENCE', "SYNC", "ASYNC"]
+            'ROUND_ROBIN', 'UNTIL', 'CONVERGENCE', "SYNC", "ASYNC",
+            'ALTER', 'PULL_IDB', 'PULL_EDB', 'BUILD_EDB']
```
senderista: I'm slightly concerned about the `ALTER` keyword causing misunderstanding (that it means "alter" and not "alternate"). Maybe we should ask some people unfamiliar with this syntax what it suggests to them. An unambiguous *cough* alternative might be `ALTERN` or just `ALTERNATE`.

raco/myrial/optimizer_tests.py
```diff
@@ -1325,18 +1331,18 @@ def test_galaxy_evolution(self):
...
+           emit Galaxies.time+1, Edges.gid2 as gid];
+        until convergence async build_EDB;
         store(Galaxies, Galaxies);
         """
         lp = self.get_logical_plan(query, async_ft='REJOIN')
```
senderista: I can't comment on the definitions of `get_logical_plan()`, `logical_to_physical()`, `compile_to_json()` here, so: don't we need to add a checkbox to MyriaWeb exposing the `async_ft` option? I can create an issue in the `myria-web` repo if so.

raco/myrial/scanner.py
BrandonHaynes: +1 `ALTERNATE`

raco/myrial/optimizer_tests.py
jingjingwang: Yes we need it, and I have a `myria-web` branch. Will link to it later.

raco/algebra.py
```diff
@@ -66,6 +70,8 @@ def scheme(self):
     def walk(self):
         """Return an iterator over the tree of operators."""
         yield self
+        if self.stop_recursion:
```
jingjingwang: Yep they need it as well. One concern about adding this `stop_recursion` is that I'm not sure if I have added it to every necessary place. Right now I'm only relying on tests.

raco/myrial/optimizer_tests.py
jingjingwang: Yay!

raco/backends/myria/myria.py
jingjingwang: Turned out that `preorder` doesn't work if `stop_recursion` is added to it, since finding an `MyriaIDBController` requires to pass a consumer with `stop_recursion=True`.

jingjingwang: Exactly, I can't remember why I came up with such a weird name...

jingjingwang:
Goes with https://github.com/uwescience/myria/pull/877.

PR #546, opened 2017-01-04
Track relations in query scan

### LINKED ISSUE #495: DBQueryScan operator includes source relations
see https://github.com/uwescience/myria/issues/795

senderista:
fixes #495

This is a prerequisite for pushing debroadcast into MyriaX as @jingjingwang and I discussed after the 1/3/2016 Myria meeting.

### DIFF
--- a/raco/backends/myria/myria.py
+++ b/raco/backends/myria/myria.py
@@ -920,10 +920,12 @@ class MyriaQueryScan(algebra.ZeroaryOperator, MyriaOperator):
 
     """A Myria Query Scan"""
 
-    def __init__(self, sql, scheme, num_tuples=algebra.DEFAULT_CARDINALITY,
+    def __init__(self, sql, scheme, source_relation_keys,
+                 num_tuples=algebra.DEFAULT_CARDINALITY,
                  partitioning=RepresentationProperties()):
         self.sql = str(sql)
         self._scheme = scheme
+        self.source_relation_keys = source_relation_keys
         self._num_tuples = num_tuples
         self._partitioning = partitioning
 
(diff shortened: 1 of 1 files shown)


raco/backends/myria/myria.py
```diff
@@ -920,10 +920,12 @@ class MyriaQueryScan(algebra.ZeroaryOperator, MyriaOperator):
 
     """A Myria Query Scan"""
 
-    def __init__(self, sql, scheme, num_tuples=algebra.DEFAULT_CARDINALITY,
+    def __init__(self, sql, scheme, source_relation_keys,
+                 num_tuples=algebra.DEFAULT_CARDINALITY,
```
jingjingwang: Do we also need to change its `__repr__()`? (My understanding is that some tests may fail if `__repr__()` is inconsistent with the constructor..?)

senderista: The current `__repr__` implementation is consistent with the constructor:

```
def __repr__(self):
        return ("{op}({sql!r}, {sch!r}, {rels!r}, {nt!r}, {part!r}, {db!r})"
                .format(op=self.opname(), sql=self.sql, sch=self._scheme,
                        rels=self.source_relation_keys, nt=self._num_tuples,
                        part=self._partitioning, db=self._debroadcast))
```

senderista: Actually, I copied that from the `add_broadcast` branch. I put this in a separate PR because the changes were self-contained WRT broadcast, but maybe you should just review that branch instead and get all the changes at once. I'll close the existing RACO broadcast PR (since it's on an obsolete branch) and open a new one.

senderista:
Closing since this branch has already been merged into the `add_broadcast` branch.

PR #548, opened 2017-01-05
Myria broadcast partitioning support

senderista:
This supersedes https://github.com/uwescience/raco/pull/517. I also fixed another issue in this PR (https://github.com/uwescience/raco/issues/495) since it needed to be done to properly support debroadcast of QueryScans.


raco/algebra.py
```diff
@@ -1525,25 +1537,35 @@ def __hash__(self):
         return ("%s-%s" % (self.opname(), self.relation_key)).__hash__()
 
     def shortStr(self):
-        return "%s(%s)" % (self.opname(), self.relation_key)
+        return "%s(%s, %s)" % (self.opname(),
+                               self.relation_key,
+                               "debroad" if self._debroadcast else "")
```
shrjain: Nitpick: I don't like this. An alternate I can think of is to have BroadcastScan(<<self. relation_key>>) and DeBroadcastScan(<<self. relation_key>>), but thats not pretty either.

raco/algebra.py
```diff
@@ -1665,9 +1687,10 @@ class ScanTemp(ZeroaryOperator):
 
     """Read the contents of a temporary relation."""
 
-    def __init__(self, name=None, _scheme=None):
+    def __init__(self, name=None, _scheme=None, debroadcast=False):
```
shrjain: Another nitpick: 
suggest to change _scheme in the arglist to scheme.

raco/backends/myria/myria.py
```diff
@@ -1195,13 +1215,19 @@ def fire(self, expr):
         right_cols = [expression.UnnamedAttributeRef(i)
                       for i in right_cols]
 
+        print "SHUFFLE BEFORE JOIN ", expr
```
shrjain: Is this a debug statement?

shrjain:
Mostly nitpicks, otherwise looks good.

raco/backends/myria/myria.py
senderista: good catch, I think @bmyerz left it in :)

raco/algebra.py
senderista: done

raco/algebra.py
senderista: I don't like it either, it's what I inherited from the first version and I can't think of anything better. Note that something like `BroadcastScan` for `debroadcast==False` doesn't really make sense because `debroadcast==False` does not imply that the relation is broadcast. I think a logical `DeBroadcast` operator (which I removed from the first version) would make the visualized query plan clearer, but maybe the next best thing is `DeBroadcast(Scan(<<self. relation_key>>))`?

senderista:
@shrjain if you wanna take a look at the new `ShuffleAfterFileScan` rule, that would be swell :)

shrjain:
@senderista: But from the mail thread I gather that this won't work without pushselects rule being updated.

jingjingwang:
Will the push-select rule be updated in this PR?

senderista:
I guess we may as well implement it, it shouldn't be hard...

raco/backends/myria/myria.py
```diff
@@ -1785,6 +1807,8 @@ def opt_rules(self, **kwargs):
             rules.push_select,
             rules.push_project,
             rules.push_apply,
+            [PushSelectThroughShuffle()],
+            rules.push_select,
```
shrjain: This rule seems to appear twice.

raco/backends/myria/myria.py
```diff
@@ -1852,6 +1876,8 @@ def opt_rules(self, **kwargs):
             rules.push_project,
             merge_to_nary_join,
             rules.push_apply,
+            [PushSelectThroughShuffle()],
+            rules.push_select,
```
shrjain: Same comment as above. This rule seems to appear twice.

raco/backends/myria/myria.py
shrjain: I mean rules.push_select is on line 1807 and also on 1811.

senderista: That is intentional. Since the existing `PushSelects` rule cannot push selects below a shuffle, it has no opportunity to perform optimizations below the shuffle (e.g., pushing the select down as far as possible, merging with other selects). Therefore I need to run the rule (actually the set of rules in `raco.rules.push_select`) again to ensure that selects just pushed below a shuffle are fully optimized. There are several other examples of this in RACO, e.g., the definition of `push_apply` in `raco/raco/rules.py`:

```
push_apply = [
    # These really ought to be run until convergence.
    # For now, run twice and finish with PushApply.
    PushApply(),
    RemoveUnusedColumns(),
    PushApply(),
    RemoveUnusedColumns(),
    PushApply(),
    RemoveNoOpApply(),
]
```

shrjain: Ah. Got it.

senderista:
@jingjingwang @shrjain I think this should be ready to merge if you can make a final pass. I've added the new `PushSelectThroughShuffle` rule and some minimal test coverage for the new rules (`ShuffleAfterFileScan`, `PushSelectThroughShuffle`).

bmyerz:
@senderista Glad to see this make it in!! Sorry I couldn't help push it through.

PR #558, opened 2017-05-03
do not regenerate op ids

jingjingwang:
This fixes the random failures of `ConnectedComponentTest` and `JoinChainAndCircleTest` -- the op_ids should be unique across the whole query.


senderista:
@BrandonHaynes could you review, since it looks like  you added the reinitialization of `op_ids` in `compile_fragment()` that this change removes?

senderista:
NVM, I see this was the result of a bad merge; the reinitialization was already removed in https://github.com/uwescience/raco/commit/c4db745e674d5793eed889104905d205facc4480.

PR #559, opened 2017-05-09
Fix arguments to variadic (Python) functions

BrandonHaynes:
Prior to this fix, only positional arguments were allowed.  This PR correctly maps arbitrary argument expressions to UDF parameters.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
   - "2.7"
 install:
+  - pip install setuptools --upgrade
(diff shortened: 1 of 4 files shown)


BrandonHaynes:
@parmitam why are we adding commits unrelated to the fix in this PR?

parmitam:
I can't test UDAs (or complex UDFs) without SQL supporting the blob type.

parmitam:
This looks great, ready to checkin.

PR #560, opened 2017-05-10
Fluent udf final

BrandonHaynes:


BrandonHaynes:
Created PR in wrong repository :-\

Issue #555, opened 2017-03-06
Support BLOB literals in MyriaL


senderista:
This would be useful for tests, if not for real-world applications. Presumably the literal format should be base64 strings.

senderista:
@BrandonHaynes: "just steal the Python 3 syntax for byte literals `b'\xbe\xef'` and have separate base64 encode/decode functions"

senderista:
fixed in #562

PR #562, opened 2017-05-10
Blob literal: adding blob literal to support BLOB data type in Myria

parmitam:


parmitam:
I have a long string of changes here, please let me know if you need me to sqash for review.
Overview:
- new literal type BLOB_LITERAL
- new token NULL
- only value the new BLOB_LITERAL can be be assigned is NULL
- this is required for the janino  compiler to be able to compile the init expression containing BLOB_TYPE.

parmitam:
Any comments?
Can we push this through?

BrandonHaynes:
This implementation conflates NULL with a zero-length byte array.  I'm not in favor of introducing nullability into Myria as a special case for one attribute type.

parmitam:
I am not sure I understand the issue with allowing a blob-type being null. Could you help me understand why this is a problem?

raco/myrial/parser.py
```diff
@@ -975,8 +975,8 @@ def p_sexpr_boolean_literal(p):
 
     @staticmethod
     def p_sexpr_blob_literal(p):
-        '''sexpr : NULL'''
-        p[0] = sexpr.BlobLiteral(p[1])
+        '''sexpr : EMPTY_BLOB'''
```
senderista: Um, I don't see where `EMPTY_BLOB` is defined. If it's not a keyword like `NULL`, what is it?

senderista: I see that Travis failed for this exact reason...

senderista:
I now think we should just use the empty blob literal `b''` (assuming we've settled on the Python-style byte literal syntax described in [PEP-3112](https://www.python.org/dev/peps/pep-3112/)), and add a check to `compile_expr()` that `op.value` is the empty literal (until we add general blob literal support). That seems to me to be the cleanest and most forward-compatible solution (i.e., assuming that we will eventually support blob literals as the value of a blob type anywhere in MyriaL).

parmitam:
How would you distinguish between b'' and empty string in python?

BrandonHaynes:
Slowly starting to catch up on my backlog :)

Agree with @senderista that `b''` is the best literal syntax here, and that it's fine to not (yet) support arbitrary byte string literals.  I'd also be comfortable with an implicit conversion from `''` to `b''`, but am not sure how much work that would involve.  (Assuming that this is still an open discussion, isn't it possible to write a UDA that emits its initial null state?  If a null value ever leaks into a relation, we are then forced to make operators null-aware.)

Assuming we want to live in a Python 2 universe, we will probably want to use the `unicode` type for strings and `bytes` for byte strings.  `basestring` and `bytearray` is another option, but I prefer that the byte strings be immutable.

parmitam:
there seem to be two issues here:
a) value null of the variable ending up in the relation 
b) using a keyword to specify and empty bytearray/bytes/b'' which is indistinguishable from an empty string when expressed in MyriaL.
For a) I'll change it so that we never send null to the relation -- @BrandonHaynes  is right, it could end up in a relation and we might need to change more to ensure everything can handle null values.
For b) I am not sure that we can do this without a keyword. 
if we are agreed that we can leave the arbitrary byte string literals for later, I'll make  the change specified in a) and  get this change in for now.

BrandonHaynes:
I don't understand why we cannot distinguish `b''` from `''` in MyriaL.

parmitam:
b'' parses as empty string _LexToken(STRING_LITERAL,'',3,38)_
a regular expression specified to look for a prepended b causes errors in lexer as well.
If you have better ideas, lmk.

BrandonHaynes:
I implemented byte string parsing using the `b''` syntax.  Do you want me to push it to a separate branch?

senderista:
Let me know what you think about the `b'\xFF'` format I implemented. It's obviously more verbose than just using the unescaped hex codes like `b'FF'`.

BrandonHaynes:
Yeah, this is essentially identical to my implementation.  Might want to support double-quotes as well since ordinary strings do.

parmitam:
This works end-to-end.

senderista:
@parmitam @BrandonHaynes any reason not to merge this (and the MyriaX `blob_literal` branch) into master? (I still have to fix the new integration test for `BITSET` expressions in MyriaX, but I guess I can leave that test out for now.)

raco/myrial/parser.py
```diff
@@ -1234,6 +1241,8 @@ def parse(self, s, udas=None):
     @staticmethod
     def p_error(token):
         if token:
+            print "error causing token"
```
BrandonHaynes: Remove debugging printfs.

BrandonHaynes:
Looks fine to me, modulo the stray debugging print statements.

senderista:
fixes https://github.com/uwescience/raco/issues/555

PR #563, opened 2017-05-12
Variadic argument fix

BrandonHaynes:
Some of these fixes were missing from the fluent UDF PR.


parmitam:
Any issue with merging this into master?

PR #564, opened 2017-06-07
Encoding for CSVFragmentTupleSource

jortiz16:
Depends on https://github.com/uwescience/myria/pull/897

Added CSVFragmentTupleSource encoding as the default when loading CSV data from S3.

### DIFF
--- a/raco/backends/myria/myria.py
+++ b/raco/backends/myria/myria.py
@@ -19,6 +19,7 @@ from raco.expression import WORKERID, COUNTALL
 from raco.representation import RepresentationProperties
 from raco.rules import distributed_group_by, check_partition_equality
 from raco.expression import util
+from urlparse import urlparse
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


raco/backends/myria/myria.py
```diff
@@ -210,17 +211,31 @@ def compileme(self):
...
-                "opType": "TupleSource",
-                "reader": dict({
-                    "readerType": "CSV",
-                    "schema": scheme_to_schema(self.scheme())
-                }, **self.options),
-                "source": {
-                    "dataType": "URI",
-                    "uri": self.path
+            url_parser = urlparse(self.path)
+            if(url_parser.scheme == 's3'):
```
senderista: This is not idomatic Python; it should just be
```
if url_parser.scheme == 's3':
```

raco/backends/myria/myria.py
```diff
@@ -210,17 +211,31 @@ def compileme(self):
...
-            encoding = {
-                "opType": "TupleSource",
-                "reader": dict({
-                    "readerType": "CSV",
-                    "schema": scheme_to_schema(self.scheme())
-                }, **self.options),
-                "source": {
-                    "dataType": "URI",
-                    "uri": self.path
+            url_parser = urlparse(self.path)
```
senderista: This is a misleading variable name; a better name would be `parsed_url`.

PR #565, opened 2017-07-20
force ordering of partitioning attributes

### LINKED ISSUE #515: Order ignored in partition attributes
Thanks to @jingjingwang and @shbae 

...

> I guess it's caused by the new RACO optimization where a shuffle is eliminated if the relation is already partitioned in the same way, but the order of the hash indices seems to be incorrect. This is the join of the JSON plan:
> 
> ```
>     "operators": [
>       {
>         "opName": "MyriaScan(shbae:graphs:toy_graph2)",
>         "opId": 18,
>         "relationKey": {
>           "relationName": "toy_graph2",
>           "userName": "shbae",
>           "programName": "graphs"
>         },
>         "opType": "TableScan"
>       },
>       {
>         "opName": "MyriaShuffleConsumer",
>         "opId": 19,
>         "argOperatorId": 20,
>         "opType": "ShuffleConsumer"
>       },
>       {
>         "argColumnNames": [
>           "row",
>           "value",
>           "value1",
>           "col",
>           "value11"
>         ],
>         "opName": "MyriaSymmetricHashJoin((($5 = $0) and ($2 = $4)); $0,$1,$3,$5,$6)",
>         "opId": 21,
>         "argChild2": "18",
>         "argChild1": "19",
>         "argSelect1": [
>           0,
>           1,
>           3
>         ],
>         "argSelect2": [
>           1,
>           2
>         ],
>         "argColumns1": [
>           0,
>           2
>         ],
>         "argColumns2": [
>           1,
>           0
>         ],
>         "opType": "SymmetricHashJoin"
>       },
> ```
> 
> operator 18 should give us tuples partitioned by ($1, $0) as specified in argColumns2, but shbae:graphs:toy_graph2 was partitioned using ($0, $1) (http://ec2-54-183-236-108.us-west-1.compute.amazonaws.com:8753/dataset/user-shbae/program-graphs/relation-toy_graph2), so there still needs to be a Shuffle. I used round-robin in my deployment so my plan had another ShuffleConsumer instead of a Scan, so I had no problem. 

I think the problem is that hash partitioning in Raco is using a set instead of a list. It is ignoring order but shouldn't be. Should be an easy fix.

senderista:
Fixes #515. Unfortunately I can't add a new RACO test to verify this fixes the bug, since FakeDB doesn't support physical representation properties (this is mentioned in https://github.com/uwescience/raco/issues/511), so I added a new integration test to MyriaX (which I'll merge as soon as this PR is merged): https://github.com/uwescience/myria/commit/b4f63165299a062081f5f730ba1becb6fde79e17

### DIFF
--- a/raco/algebra.py
+++ b/raco/algebra.py
@@ -675,10 +675,10 @@ class CompositeBinaryOperator(BinaryOperator):
 
         # TODO: being conservative, just take one side because
         # TODO: we don't support conjuncs in partition info
-        if self.left.partitioning().hash_partitioned != frozenset():
+        if self.left.partitioning().hash_partitioned != tuple():
             return RepresentationProperties(
                 hash_partitioned=self.left.partitioning().hash_partitioned)
-        elif self.right.partitioning().hash_partitioned != frozenset():
+        elif self.right.partitioning().hash_partitioned != tuple():
             return RepresentationProperties(
                 hash_partitioned=self.right.partitioning().hash_partitioned)
         elif self.left.partitioning().broadcasted and (
@@ -785,7 +785,7 @@ def project_partitioning(columnlist, input_partitioning):
     """Return the partitioning for a simple projection that supports
     duplicates, swapping, and removal"""
 
-    if input_partitioning.hash_partitioned <= frozenset(columnlist):
+    if set(input_partitioning.hash_partitioned) <= set(columnlist):
         # Translate to new schema.
 
         # In general, Apply can make hash partitioning into a disjunction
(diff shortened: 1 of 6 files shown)


senderista:
There's still an open optimization issue (although not a correctness issue). Reordering the join conditions will no longer produce an incorrect query plan, but it could force an unnecessary shuffle (which happens in the referenced integration test). When one side of a join is already partitioned on the join attributes, and the other input is not partitioned, the generated shuffle for the unpartitioned input should ensure its partitioning order is compatible with the already-partitioned input. If we allow the order of join conditions to determine the order of partitioning attributes (which we do now), then this will generally not be the case, and we will generate an unnecessary shuffle for the already-partitioned input. Since the ordering of partitioning attributes has no user-visible significance, there is no reason to let it be determined by the order of conditions in a query.

senderista:
Opened https://github.com/uwescience/raco/issues/566 for the issue in my comment.

senderista:
MyriaX integration test PR: https://github.com/uwescience/myria/pull/905

PR #567, opened 2017-07-20
Fix for myria-python #67

BrandonHaynes:


BrandonHaynes:
@senderista any objections to merging this?

raco/myrial/parser.py
```diff
@@ -297,7 +297,8 @@ def add_python_udf(name, typ, **kwargs):
         :param typ: The output type of the function
         :type typ: string
         """
-        if name in Parser.udf_functions:
+        if name in Parser.udf_functions and \
+                not kwargs.pop('overwrite_if_exists', False):
```
senderista: why use `pop()` here?

BrandonHaynes: Other `kwargs` are sent (see https://github.com/uwescience/raco/blob/6f3e834b484acef1757910af974aa096543901b6/raco/myrial/parser.py#L304) to the `VariadicFunction` constructor (and ultimately used in the `Function` instantiation when bound); neither `VariadicFunction` nor `Function` know anything about overwriting registered functions, and may choke on an unexpected keyword argument.

I could make the parameter explicit like I did in `add_udf` if you think it would be more clear.

senderista: Yeah, I think it'd be clearer to reserve `kwargs` for the final `VariadicFunction`.

BrandonHaynes: Agree -- done!

PR #568, opened 2017-07-21
adding support for OrderBy and Limit in MyriaL

jortiz16:
* added support for OrderBy in Myrial (but must be paired with a Limit operator). Can support queries with this syntax:

```
T1 = [from scan(TwitterK) as t ORDERBY $0 ASC, $1 DESC emit *];
T2 = LIMIT(T1,50);
store(T2, TwitterKOrdered);)
```

I wasn't sure if we wanted to use the `ASC` or `DESC` keywords in MyriaL, but I added them there for now. If neither `ASC` or `DESC` are described, we order ascending by default.

* added new rule to push a Collect operator below the OrderBy

### DIFF
--- a/raco/algebra.py
+++ b/raco/algebra.py
@@ -1237,8 +1237,8 @@ class OrderBy(UnaryOperator):
         return "{op}({inp!r}, {scol!r}, {asc!r})".format(
             op=self.opname(),
             inp=self.input,
-            scol=list(self.sort_columns),
-            asc=list(self.ascending))
+            scol=self.sort_columns,
+            asc=self.ascending)
 
     def num_tuples(self):
         return self.input.num_tuples()
--- a/raco/algebra.py
+++ b/raco/algebra.py
@@ -1221,13 +1221,24 @@ class GroupBy(UnaryOperator):
 
 class OrderBy(UnaryOperator):
 
-    """ Logical Sort operator
-    """
+    """ Logical OrderBy operator"""
 
     def __init__(self, input=None, sort_columns=None, ascending=None):
         UnaryOperator.__init__(self, input)
         self.sort_columns = sort_columns
         self.ascending = ascending
+
+    def __eq__(self, other):
+        return UnaryOperator.__eq__(self, other) and \
+            self.sort_columns == other.sort_columns and \
+            self.ascending == other.ascending
+
+    def __repr__(self):
+        return "{op}({inp!r}, {scol!r}, {asc!r})".format(
+            op=self.opname(),
+            inp=self.input,
+            scol=list(self.sort_columns),
+            asc=list(self.ascending))
 
     def num_tuples(self):
         return self.input.num_tuples()
--- a/raco/backends/myria/myria.py
+++ b/raco/backends/myria/myria.py
@@ -1270,13 +1270,16 @@ class CollectBeforeLimit(rules.Rule):
                                   MyriaLimit(exp.count, exp.input)))
         return exp
 
+
 class GlobalOrderBy(rules.Rule):
 
     def fire(self, exp):
         if exp.__class__ == algebra.OrderBy:
-            return MyriaInMemoryOrderBy(algebra.Collect(
-                                            exp.input),exp.sort_columns, exp.ascending)
+            return MyriaInMemoryOrderBy(
+                algebra.Collect(exp.input),
+                exp.sort_columns, exp.ascending)
         return exp
+
 
 class ShuffleBeforeSetop(rules.Rule):
 
@@ -2445,7 +2448,9 @@ def compile_to_json(raw_query, logical_plan, physical_plan,
     assert isinstance(physical_plan, subplan_ops), \
         'Physical plan must be a subplan operator, not {}'.format(type(physical_plan))  # noqa
 
-    if('OrderBy' in str(logical_plan) and 'Limit' not in str(logical_plan)):
+    if('OrderBy' in str(logical_plan)
+            and 'Limit' not in str(logical_plan)
+            and 'MyriaHyperCube' not in str(logical_plan)):
         raise Exception("OrderBy queries must use the Limit operator")
 
     # raw_query must be a string
--- a/raco/backends/myria/myria.py
+++ b/raco/backends/myria/myria.py
@@ -2433,8 +2433,10 @@ def compile_to_json(raw_query, logical_plan, physical_plan,
     # Store/StoreTemp is a reasonable physical plan... for now.
     root_ops = (algebra.Store, algebra.StoreTemp, algebra.Sink)
     if isinstance(physical_plan, root_ops):
-        if isinstance(logical_plan.input, algebra.OrderBy):
-            raise Exception("Order By queries must use the Limit operator")
+        if isinstance(physical_plan.input, MyriaSplitConsumer):
+            splitOperator = physical_plan.input.input.input
+            if isinstance(splitOperator, MyriaInMemoryOrderBy):
+                raise Exception("Order By queries must use the Limit operator")
         physical_plan = algebra.Parallel([physical_plan])
 
     subplan_ops = (algebra.Parallel, algebra.Sequence, algebra.DoWhile,
--- a/raco/backends/myria/myria.py
+++ b/raco/backends/myria/myria.py
@@ -2433,10 +2433,6 @@ def compile_to_json(raw_query, logical_plan, physical_plan,
     # Store/StoreTemp is a reasonable physical plan... for now.
     root_ops = (algebra.Store, algebra.StoreTemp, algebra.Sink)
     if isinstance(physical_plan, root_ops):
-        if isinstance(physical_plan.input, MyriaSplitConsumer):
-            splitOperator = physical_plan.input.input.input
-            if isinstance(splitOperator, MyriaInMemoryOrderBy):
-                raise Exception("Order By queries must use the Limit operator")
         physical_plan = algebra.Parallel([physical_plan])
 
     subplan_ops = (algebra.Parallel, algebra.Sequence, algebra.DoWhile,
--- a/raco/backends/myria/myria.py
+++ b/raco/backends/myria/myria.py
@@ -1268,9 +1268,15 @@ class CollectBeforeLimit(rules.Rule):
             return MyriaLimit(exp.count,
                               algebra.Collect(
                                   MyriaLimit(exp.count, exp.input)))
-
         return exp
 
+class GlobalOrderBy(rules.Rule):
+
+    def fire(self, exp):
+        if exp.__class__ == algebra.OrderBy:
+            return MyriaInMemoryOrderBy(algebra.Collect(
+                                            exp.input),exp.sort_columns, exp.ascending)
+        return exp
 
 class ShuffleBeforeSetop(rules.Rule):
 
@@ -1844,6 +1850,7 @@ left_deep_tree_shuffle_logic = [
     ShuffleBeforeIDBController(),
     BroadcastBeforeCross(),
     ShuffleAfterSingleton(),
+    GlobalOrderBy(),
     CollectBeforeLimit(),
     ShuffleAfterFileScan(),
 ]
@@ -1872,6 +1879,7 @@ myriafy = [
     rules.OneToOne(algebra.UnionAll, MyriaUnionAll),
     rules.OneToOne(algebra.Difference, MyriaDifference),
     rules.OneToOne(algebra.OrderBy, MyriaInMemoryOrderBy),
+    rules.OneToOne(algebra.Limit, MyriaLimit),
     rules.OneToOne(algebra.Sink, MyriaSink),
     rules.OneToOne(algebra.IDBController, MyriaIDBController),
 ]
@@ -2436,6 +2444,9 @@ def compile_to_json(raw_query, logical_plan, physical_plan,
                    algebra.UntilConvergence)
     assert isinstance(physical_plan, subplan_ops), \
         'Physical plan must be a subplan operator, not {}'.format(type(physical_plan))  # noqa
+
+    if('OrderBy' in str(logical_plan) and 'Limit' not in str(logical_plan)):
+        raise Exception("OrderBy queries must use the Limit operator")
 
     # raw_query must be a string
     if not isinstance(raw_query, basestring):
--- a/raco/backends/myria/myria.py
+++ b/raco/backends/myria/myria.py
@@ -1258,26 +1258,19 @@ class BreakSplit(rules.Rule):
         return consumer
 
 
-class CollectBeforeLimit(rules.Rule):
-
-    """Similar to a decomposable GroupBy, rewrite Limit as
-    Limit[Collect[Limit]]"""
+class LimitOrderBy(rules.Rule):
 
     def fire(self, exp):
-        if exp.__class__ == algebra.Limit:
-            return MyriaLimit(exp.count,
-                              algebra.Collect(
-                                  MyriaLimit(exp.count, exp.input)))
-        return exp
-
-
-class GlobalOrderBy(rules.Rule):
-
-    def fire(self, exp):
-        if exp.__class__ == algebra.OrderBy:
-            return MyriaInMemoryOrderBy(
-                algebra.Collect(exp.input),
-                exp.sort_columns, exp.ascending)
+        if exp.__class__ == algebra.Limit and \
+                isinstance(exp.children()[0], algebra.OrderBy):
+            child = exp.children()[0]
+            return MyriaLimit(exp.count, MyriaInMemoryOrderBy(
+                algebra.Collect(
+                    MyriaLimit(exp.count,
+                               MyriaInMemoryOrderBy(child.input,
+                                                    child.sort_columns,
+                                                    child.ascending))),
+                child.sort_columns, child.ascending))
         return exp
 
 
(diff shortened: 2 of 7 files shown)


raco/myrial/parser.py
```diff
@@ -788,6 +789,36 @@ def p_opt_where_clause(p):
...
+    def p_explicit_orderby_list(p):
+        """orderby_arg_list : orderby_arg_list COMMA orderby_arg
+                            | orderby_arg"""
+        if len(p) == 4:
+            p[0] = p[1] + (p[3],)
+        else:
+            p[0] = (p[1],)
+
+    @staticmethod
+    def p_explicit_orderby_explicit(p):
```
senderista: Shouldn't this be `p_explicit_orderby_arg`?

raco/myrial/interpreter.py
```diff
@@ -232,6 +234,13 @@ def bagcomp(self, from_clause, where_clause, emit_clause):
             where_clause.typeof(op.scheme(), None)
             op = raco.algebra.Select(condition=where_clause, input=op)
 
+        if orderby_clause:
+            orderbyTuples = zip(*orderby_clause)
+            if orderby_clause:
```
senderista: why the duplicate conditional?

raco/myrial/interpreter.py
```diff
@@ -232,6 +234,13 @@ def bagcomp(self, from_clause, where_clause, emit_clause):
             where_clause.typeof(op.scheme(), None)
             op = raco.algebra.Select(condition=where_clause, input=op)
 
+        if orderby_clause:
+            orderbyTuples = zip(*orderby_clause)
```
senderista: Now that is cool!

raco/myrial/scanner.py
```diff
@@ -9,11 +9,11 @@
             'ELSE', 'END', 'CONST', 'LOAD', 'DUMP', 'CSV', 'SCHEMA',
             'OPP', 'TIPSY', 'UDA', 'TRUE', 'FALSE', 'HASH', 'BROADCAST',
             'ROUND_ROBIN', 'UNTIL', 'CONVERGENCE', "SYNC", "ASYNC",
-            'ALTERNATE', 'PULL_IDB', 'PULL_EDB', 'BUILD_EDB']
+            'ALTERNATE', 'PULL_IDB', 'PULL_EDB', 'BUILD_EDB', 'ASC', 'DESC']
 
 types = ['INT', 'STRING', 'FLOAT', 'BOOLEAN', 'BLOB']
 
-comprehension_keywords = ['SELECT', 'AS', 'EMIT', 'FROM', 'WHERE']
+comprehension_keywords = ['SELECT', 'AS', 'EMIT', 'FROM', 'WHERE', 'ORDERBY']
```
senderista: I think it's important to use the conventional syntax `ORDER BY`, even if it's a pain to deal with the embedded space in the token. Here's some advice on that: https://stackoverflow.com/questions/26596068/identifying-keyword-pairs-in-lex

raco/myrial/interpreter.py
```diff
@@ -145,7 +145,7 @@ def empty(_scheme):
 
     def select(self, args):
         """Evaluate a select-from-where expression."""
-        op = self.bagcomp(args.from_, args.where, args.select)
+        op = self.bagcomp(args.from_, args.where, None, args.select)
```
senderista: Don't we want to allow `ORDER BY` in SQL `SELECT` statements, not just MyriaL bag comprehensions?

raco/backends/myria/myria.py
```diff
@@ -1872,6 +1882,7 @@ def fire(self, expr):
     rules.OneToOne(algebra.UnionAll, MyriaUnionAll),
     rules.OneToOne(algebra.Difference, MyriaDifference),
     rules.OneToOne(algebra.OrderBy, MyriaInMemoryOrderBy),
+    rules.OneToOne(algebra.Limit, MyriaLimit),
```
senderista: As I mentioned, we should have a 1:1 rule for `OrderBy` as well.

raco/backends/myria/myria.py
```diff
@@ -1268,7 +1268,16 @@ def fire(self, exp):
...
                               algebra.Collect(
                                   MyriaLimit(exp.count, exp.input)))
+        return exp
+
+
+class GlobalOrderBy(rules.Rule):
 
+    def fire(self, exp):
+        if exp.__class__ == algebra.OrderBy:
+            return MyriaInMemoryOrderBy(
```
jortiz16: If we add a local OrderBy, the Collect operator will gather tuples in no particular order, right? I'm not sure I understand how this would help. I can certainly add it though.

raco/backends/myria/myria.py
jortiz16: The 1:1 rule for the OrderBy was already on the list :)

raco/backends/myria/myria.py
```diff
@@ -2437,6 +2448,11 @@ def compile_to_json(raw_query, logical_plan, physical_plan,
     assert isinstance(physical_plan, subplan_ops), \
         'Physical plan must be a subplan operator, not {}'.format(type(physical_plan))  # noqa
 
+    if('OrderBy' in str(logical_plan)
```
jortiz16: Sounds good! Makes sense.

raco/myrial/parser.py
jortiz16: yes

raco/backends/myria/myria.py
senderista: The order in which the `Collect` operator receives tuples is immaterial. The local `OrderBy` is just there in order to apply `Limit`. The only user-visible ordering takes place in the final, global `OrderBy` on the coordinator.

Actually, I'm not sure now that either the existing `CollectBeforeLimit` or the new `GlobalOrderBy` rules will properly transform a `Limit(OrderBy)` logical expression. The transformation should be `Limit(OrderBy) ->  Limit(OrderBy(Collect(Limit(OrderBy))))`, but `CollectBeforeLimit` would transform it into `Limit(Collect(Limit(OrderBy))` (which is clearly incorrect), and your version of `GlobalOrderBy` (along with `CollectBeforeLimit`, which is evaluated first) would transform it like `Limit(OrderBy) -> Limit(Collect(Limit(OrderBy)) -> Limit(Collect(Limit(OrderBy(Collect(input)))))`, which would presumably be collapsed into `Limit(OrderBy(Collect(input)))`, since consecutive `Limit`s and `Collect`s are no-ops. This final transformation is actually correct, but highly suboptimal, since it omits the local `Limit` and `OrderBy`, forcing the entire original input to be streamed to the coordinator, which defeats the purpose of these local/global decomposition rules.

senderista: You may actually need a special rule that recognizes just the `Limit(OrderBy)` pattern (there's no point in having a rule `OrderBy -> OrderBy(Collect(OrderBy))`, although if we had a `Merge` exchange operator it would be sensible to have a rule `OrderBy -> Merge(OrderBy)`). Also the `CollectBeforeLimit` rule may need to be modified to ignore this pattern (I think it's really useless anyway since there's not much point in using `LIMIT` without `ORDER BY`).

raco/backends/myria/myria.py
jortiz16: I tried switching the CollectBeforeLimit rule to use the 'algebra.Limit' operator rule, but it leads to infinite recursion when applying the rules. It seems like applying the physical operator was purposely used as a stopping condition (so the rules don't apply to the new local limit operator we generate).

senderista: Makes sense. As I mentioned, we probably don't even need this rule anyway.

raco/backends/myria/myria.py
jortiz16: **1)** I implemented this approach (checking whether the Root has an Order By child), but I'm not sure it would always work. For example, for this query:

```
T1 = [from scan(TwitterK) as t ORDERBY $0 ASC, $1 DESC emit $0];  store(T1, TwitterKOrdered);
```

we would have the following logical plan (where the Apply operator ends up being the child of the store): 
`Store(public:adhoc:TwitterKOrdered)[Apply(a=$0)[OrderBy(0+,1-)[Scan(public:adhoc:TwitterK)]]]`

**2)** There's also this other case I'm testing with:
```
T1 = [from scan(TwitterK) as t ORDERBY $0 ASC, $1 DESC emit $0, count(*)];
T2 = LIMIT(T1, 10);
store(T2, TwitterKOrdered);
```
I don't remember if we wanted to support OrderBy with aggregates, but I don't think our rules would apply here. We would need a plan like this:
`store -> limit(global) -> aggregate -> orderby (global) -> collect ->scan`

I don't believe it would be correct to apply the limit and orderby locally for this case, right?

senderista: I don't see how directly storing an `OrderBy` result makes sense. There is no ordering guarantee from a subsequent `scan()` on that stored relation (until we implement ordering metadata in the catalog and automatic index creation/`ORDER BY` SQL generation for the scan).

As for your bag comprehension example with `ORDER BY`, I think the *cough* ordering of the `ORDER BY` clause is causing confusion. I think you are inferring from the fact that the `ORDER BY` clause precedes the `EMIT` clause that the `OrderBy` operator should precede the `Apply` operator. I don't think that's the case in SQL, and I don't see why it should be the case here. I don't see why we shouldn't place the `ORDER BY` clause last in the bag comprehension syntax, making it clear that the `OrderBy` operator is the root of the operator tree (well, assuming that `LIMIT` isn't present, which it must be for now).

A sufficiently smart optimizer might push the `OrderBy` operator below an aggregate so it could use a sorted aggregate implementation, but that is orthogonal to this change (and again would require us to propagate ordering information between operators, starting at scans).

jortiz16:
@senderista Still going to fix the ORDER BY in the grammar and add LIMIT as well. Will add them soon

jortiz16:
@senderista The Travis tests aren't passing, but I'm having the same issue as the Master branch.  Do you know what the issue could be?

@jingjingwang Your recent branch is passing. Are you pushing a fix for this? 

Thanks!

jortiz16:
Any insight on the Raco/Travis issue on master I mentioned earlier? @senderista @jingjingwang

senderista:
My guess is that the Travis error is from gcc being invoked instead of clang.

senderista:
Travis is passing in master now.

jortiz16:
Great, thanks!

raco/myrial/interpreter.py
```diff
@@ -168,14 +167,17 @@ def extract_unbox_args(self, from_args, sexpr):
...
+    def bagcomp(self, from_clause, where_clause, emit_clause,
+                orderby_clause, limit_clause):
         """Evaluate a bag comprehension.
 
         from_clause: A list of tuples of the form (id, expr).  expr can
         be None, which means "read the value from the symbol table".
 
         where_clause: An optional scalar expression (raco.expression).
 
+        orderby_clause: An optional list of OrderbyArg instances.
```
senderista: need a docstring for limit_clause, also docstrings should be in same order as args

raco/myrial/interpreter.py
```diff
@@ -239,16 +241,31 @@ def bagcomp(self, from_clause, where_clause, emit_clause):
...
+                                         statemods)
         else:
             if statemods:
                 return raco.algebra.StatefulApply(emit_args, statemods, op)
             if (len(from_args) == 1 and len(emit_clause) == 1 and
                 isinstance(emit_clause[0],
                            (TableWildcardEmitArg, FullWildcardEmitArg))):
-                return op
-            return raco.algebra.Apply(emit_args, op)
+                op = op
```
senderista: what is this assignment supposed to do?

raco/myrial/interpreter.py
```diff
@@ -239,16 +241,31 @@ def bagcomp(self, from_clause, where_clause, emit_clause):
...
+            op = raco.algebra.Apply(emit_args, op)
+
+        if orderby_clause:
+            if limit_clause is None:
+                raise Exception("Order By queries must use the Limit operator")
```
senderista: let's not reference operators in parse errors, this should be something like "An ORDER BY clause must be accompanied by a LIMIT clause"

raco/myrial/interpreter.py
```diff
@@ -239,16 +241,31 @@ def bagcomp(self, from_clause, where_clause, emit_clause):
...
+            orderbyTuples = zip(*orderby_clause)
+            op = raco.algebra.OrderBy(input=op,
+                                      sort_columns=orderbyTuples[0],
+                                      ascending=orderbyTuples[1])
+
+        if limit_clause:
+            if orderby_clause is None:
+                raise Exception("Limit queries must use the Order By operator")
```
senderista: see above for better wording

raco/myrial/scanner.py
```diff
@@ -71,6 +71,14 @@
 # Regular expressions for non-trivial tokens
 
 
+def t_ORDERBY(t):
```
senderista: haha, nice approach!

raco/myrial/query_tests.py
```diff
@@ -600,23 +600,23 @@ def test_sql_repeated(self):
         expected = collections.Counter([(x[3],) for x in self.emp_table])
         self.check_result(query, expected)
 
-    def test_limit(self):
+    def test_limit_without_orderby_assert(self):
```
senderista: It's great that we have tests for the illegal combinations of LIMIT and ORDER BY, but we also need at least one test in this file that actually uses FakeDB to evaluate an ORDER BY + LIMIT query and checks the results.

raco/algebra.py
```diff
@@ -1221,14 +1221,25 @@ def __eq__(self, other):
...
+    def __eq__(self, other):
+        return UnaryOperator.__eq__(self, other) and \
+            self.sort_columns == other.sort_columns and \
+            self.ascending == other.ascending
+
+    def __repr__(self):
+        return "{op}({inp!r}, {scol!r}, {asc!r})".format(
+            op=self.opname(),
+            inp=self.input,
+            scol=list(self.sort_columns),
```
senderista: What's the motivation for copying these fields?

raco/myrial/interpreter.py
```diff
@@ -241,28 +243,29 @@ def bagcomp(self, from_clause, where_clause, emit_clause,
...
+                    "An ORDER BY clause must be accompanied by a LIMIT clause")
             orderbyTuples = zip(*orderby_clause)
             op = raco.algebra.OrderBy(input=op,
                                       sort_columns=orderbyTuples[0],
                                       ascending=orderbyTuples[1])
 
         if limit_clause:
             if orderby_clause is None:
-                raise Exception("Limit queries must use the Order By operator")
+                raise Exception(
```
senderista: Let's not raise a generic `Exception` here; I think the closest semantic match might be `InvalidStatementException`.

raco/myrial/interpreter.py
```diff
@@ -241,28 +243,29 @@ def bagcomp(self, from_clause, where_clause, emit_clause,
...
-            op = raco.algebra.Apply(emit_args, op)
+            if not (len(from_args) == 1 and len(emit_clause) == 1 and
+                    isinstance(emit_clause[0],
+                               (TableWildcardEmitArg, FullWildcardEmitArg))):
+                op = raco.algebra.Apply(emit_args, op)
 
         if orderby_clause:
             if limit_clause is None:
-                raise Exception("Order By queries must use the Limit operator")
+                raise Exception(
```
senderista: Let's not raise a generic `Exception` here; I think the closest semantic match might be `InvalidStatementException`.

raco/myrial/query_tests.py
```diff
@@ -618,6 +618,28 @@ def test_orderby_without_limit_assert(self):
         with self.assertRaises(Exception):  # noqa
             self.check_result(query, None)
 
+    def test_limit_orderby(self):
```
senderista: Both the bagcomp and SQL ORDER BY syntaxes should have multikey (with varying sort orders) tests.

raco/fakedb.py
```diff
@@ -269,6 +269,15 @@ def limit(self, op):
         it = self.evaluate(op.input)
         return itertools.islice(it, op.count)
 
+    def orderby(self, op):
+        it = self.evaluate(op.input)
+        oList = zip(op.sort_columns, op.ascending)
+        sortedList = list(it)
+        for o in oList:
```
senderista: This is the right approach for multikey sorting given a stable sort, but you need to sort in the opposite order of the given key: from least significant to most significant attribute. Therefore you just need to say
```
oList = reversed(zip(op.sort_columns, op.ascending))
```
There clearly needs to be a multikey ORDER BY test that would have caught this bug.

raco/algebra.py
jortiz16: I had trouble understanding how `__repr__` was used when compiling plans. When I remove this method, the `shortStr()` method in the operator loses state (the self.ascending and self.sort_columns are None, which results in an error).` shortStr()` is used in the `call_compile_me` method under `myria.py`.

senderista: `repr()` is used to represent query plans as executable Python code (this format is used by the `myrial` script via `plan_from_repr()`, while JSON format is used by MyriaX). Specifically, the output of `repr()` for each logical operator class must be a legal Python constructor call. Therefore, all arguments to `__init__` must appear (in the same order) as format string arguments in `__repr__`.

I think my question was unclear: what I was really asking was why you assign `scol=list(self.sort_columns)` instead of `scol=self.sort_columns`.

jortiz16: Thanks! This makes a lot more sense now. 

And to answer the other question, I do not need `list`, I should be able to remove it.

jortiz16:
sounds good, thanks!

PR #570, opened 2017-07-26
apply lists of rules recursively

jingjingwang:
An algebra was defined as a list of lists of rules but then got flattened. This limits recursive optimizations that consist of multiple rules, for example, PushApply + RemoveUnusedColumn were repeated twice, which still produces redundant columns if the query contains more than two joins. This PR removes the flattening and applied a list of rules recursively until no change. It fixes the PushApply + RemoveUnusedColumn problem and also several bugs disclosed by this change.

### DIFF
--- a/raco/backends/myria/myria.py
+++ b/raco/backends/myria/myria.py
@@ -1524,7 +1524,8 @@ class HCShuffleBeforeNaryJoin(rules.Rule):
         if not isinstance(expr, algebra.NaryJoin):
             return expr
         # check if HC shuffle has been placed before
-        shuffled_child = [isinstance(op, algebra.HyperCubeShuffle)
+        shuffled_child = [isinstance(op, algebra.HyperCubeShuffle) or
+                          isinstance(op, algebra.OrderBy)
                           for op in list(expr.children())]
         if all(shuffled_child):    # already shuffled
             assert len(expr.children()) > 0
@@ -1730,6 +1731,8 @@ class InsertSplit(rules.Rule):
         """Walk the tree starting from op and insert a split when we
         encounter a heavyweight operator."""
         if isinstance(op, MyriaAlgebra.fragment_leaves):
+            return op
+        if isinstance(op, algebra.Split):
             return op
 
         if isinstance(op, InsertSplit.heavy_ops):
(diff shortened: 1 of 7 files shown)


raco/backends/myria/myria.py
```diff
@@ -1731,6 +1732,8 @@ def insert_split_before_heavy(self, op):
         encounter a heavyweight operator."""
         if isinstance(op, MyriaAlgebra.fragment_leaves):
             return op
+        if isinstance(op, algebra.Split):
```
senderista: was this a bug?

raco/backends/myria/myria.py
```diff
@@ -1524,7 +1524,8 @@ def add_hyper_shuffle():
         if not isinstance(expr, algebra.NaryJoin):
             return expr
         # check if HC shuffle has been placed before
-        shuffled_child = [isinstance(op, algebra.HyperCubeShuffle)
+        shuffled_child = [isinstance(op, algebra.HyperCubeShuffle) or
+                          isinstance(op, algebra.OrderBy)
```
senderista: was this a bug?

senderista: wouldn't this only happen if `OrderByBeforeNaryJoin` was somehow applied before `HCShuffleBeforeNaryJoin`?

raco/backends/radish/radish.py
```diff
@@ -1638,7 +1638,7 @@ def __init__(self, array_rep, memory_scan_class=GrappaMemoryScan):
 
     def fire(self, expr):
         if isinstance(expr, algebra.Scan) \
-                and not isinstance(expr, GrappaFileScan):
+                and not isinstance(expr, cppcommon.CBaseFileScan):
```
senderista: was this a bug?

raco/backends/radish/radish.py
```diff
@@ -2314,6 +2314,8 @@ class GrappaWhileCondition(rules.Rule):
     result of the condition table"""
 
     def fire(self, expr):
+        if isinstance(expr, GrappaDoWhile):
```
senderista: was this a bug?

raco/compile.py
```diff
@@ -35,31 +37,39 @@ def write_if_enabled(self, plan, title):
...
-                          colored("  -", "red") + " %s" + "\n" +
-                          colored("  +", "green") + " %s", rule, e, newe)
-
-            newe.apply(recursiverule)
-
-            return newe
-        expr = recursiverule(expr)
-
+    for rules in rules_lists:
+        for i in range(20):
```
senderista: If this lands in an infinite recursion, it'll blow the python stack soon enough, right? Do we really need an explicit cutoff?

raco/rules.py
```diff
@@ -847,9 +833,6 @@ def __str__(self):
     # For now, run twice and finish with PushApply.
```
senderista: remove obsolete comment

raco/rules.py
```diff
@@ -557,30 +557,16 @@ def convert(n):
...
-                    # Left columns do come before right cols
-                    new_pj = algebra.ProjectingJoin(
-                        condition=child.condition, left=child.left,
-                        right=child.right, output_columns=new_cols)
-                    if new_pj.scheme() == op.scheme():
-                        return new_pj
-
             accessed = sorted(set(itertools.chain(*(accessed_columns(e)
                                                     for e in emits))))
+            new_cols = [child.output_columns[p] for p in accessed]
```
senderista: could you explain why you removed all this code?

PR #578, opened 2017-09-19
Export to S3 via MyriaL

jortiz16:
Adding the ability to export relations to S3 via MyriaL. This is not a parallel export. All workers shuffle their data to one worker before sending to S3.

By default, the relation exports in CSV format. Should we make that configurable?

Example query:

```
t = load("https://s3-us-west-2.amazonaws.com/uwdb/sampleData/TwitterK.csv", csv(schema(column0:int, column1:int), skip=1));
export(t, 's3://elastic-cluster-test/testExport');
```

### DIFF
--- a/raco/backends/myria/myria.py
+++ b/raco/backends/myria/myria.py
@@ -388,6 +388,17 @@ class MyriaSink(algebra.Sink, MyriaOperator):
         }
 
 
+class MyriaExport(algebra.Export, MyriaOperator):
+
+    def compileme(self, inputid):
+        return {
+            "opType": "TupleSink",
+            "tupleWriter": {"writerType": "CSV"},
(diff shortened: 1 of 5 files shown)


raco/backends/myria/myria.py
```diff
@@ -388,6 +388,17 @@ def compileme(self, inputid):
         }
 
 
+class MyriaExport(algebra.Export, MyriaOperator):
+
+    def compileme(self, inputid):
+        return {
+            "opType": "TupleSink",
+            "tupleWriter": {"writerType": "CSV"},
```
senderista: See below, I think we'd want to allow specifying at least the field delimiter.

senderista (dismissed):
Looks good but I'd like to discuss at least adding delimiter support.

